PR #114, opened 2026-07-26
refactor(user): Encapsulate User and PasswordHistory entity attributes

### LINKED ISSUE #109: [Refactor]: User aggregate exposes public mutable state instead of private attrs + properties
## ¿Qué parte del código se refactoriza?

`src/modules/user/domain/entities/user.py` (the `User` aggregate root) and
`src/modules/user/domain/entities/password_history.py` (`PasswordHistory`).

`User.__init__` (lines 100-117) assigns every field as a public attribute
(`self.email = email`, `self.password = password`, `self.is_admin = is_admin`,
`self.failed_login_attempts = failed_login_attempts`, etc.), with only one `@property`
(`has_password`) in the whole class. The same pattern (`self.id = ...`, etc. with 0
`@property` accessors) is present in `PasswordHistory` (lines 54-57).

This violates the Clean Architecture rule that entity attributes must be private (`_attr`)
with read-only `@property` accessors, no public mutable state. For contrast:
- `Competition` (`src/modules/competition/domain/entities/competition.py`): 15 `@property`
  accessors, 0 public attribute assignments.
- `GolfCourse`: 14 `@property` accessors.
- Other entities in the same `user` module — `UserOAuthAccount` (6 properties),
  `RefreshToken` (8 properties), `UserDevice` (9 properties) — all correctly follow the
  private-attribute + property pattern.

`User` and `PasswordHistory` are legacy outliers (User is the oldest module, v1.0.0,
predating the later DDD conventions applied to Competition/GolfCourse).

## Motivación

Found during the full backend security/architecture/DDD audit (issue #106, report at
`docs/AUDIT_BACKEND_2026-07.md`, finding H1).

Because every field on `User` is a plain public attribute, any code with a reference to a
`User` instance (a use case, an infrastructure mapper, a test helper, or a future
contributor) can write `user.is_admin = True`, `user.email = "attacker@x.com"`, or
`user.failed_login_attempts = 0` directly — bypassing every domain invariant and every
domain event the entity is supposed to emit (`UserPasswordChangedEvent`,
`AccountUnlockedEvent`, etc.).

No such bypass currently exists in the codebase — application/infrastructure code was
checked and everything goes through the entity's own methods (`update_handicap()`,
`change_password()`, ...). The risk is entirely latent: the type system provides zero
protection against a future regression, and it's inconsistent with every other entity in
the codebase.

## Riesgo / impacto en tests existentes

Medium refactor risk, well covered: 501 existing User unit tests (per CLAUDE.md) plus the
SQLAlchemy mapper for `users` (`mappers.py`) which already uses `composite()`/direct
attribute mapping for `email`/`password` (lines 157-158) — property setters need to be
preserved for imperative mapping to keep working, following the same pattern `Competition`
already uses successfully. Recommend doing this incrementally (one field group at a time)
behind the existing test suite.

## Prioridad

P1 — Alto.

agustinEDev:
## Summary

- Converts every public attribute on the `User` aggregate root and `PasswordHistory` entity to private (`_attr`) with read-only `@property` accessors, matching the pattern already established by `Competition` (15 properties, 0 public attrs) and `GolfCourse` (14 properties).
- Closes the one real encapsulation bypass found in application code: `ResendVerificationEmailUseCase` wrote directly to `user.verification_token` to persist a token generated before sending the email (needed so the token is only saved if the email send succeeds). Added `User.set_verification_token(token)` so the use case no longer touches private domain state directly.
- Updated the SQLAlchemy imperative mappers (`mappers.py`, `password_history_mapper.py`) to map table columns onto the new private attributes, following exactly the approach `Competition`'s mapper already uses successfully: scalar columns map directly to `_attr`, and the `Email`/`Password` value objects are mapped via `composite()` onto `_email`/`_password` (the attributes backing the read-only public properties), with the raw string columns behind `_email_value`/`_password_value`.
- Fixed two repositories that referenced `User.id` / `User.first_name` / `User.last_name` as SQLAlchemy class-level query expressions (`user_repository.py`, `competition_repository.py`'s `find_by_filters` join/ilike filters) — these now query via the private, SQLAlchemy-instrumented attributes (`User._id`, `User._first_name`, `User._last_name`), matching how `Competition`'s repository already queries via its own private attrs.
- Updated 5 test files that mutated `User` state directly for test setup (token-expiry simulation, account-lockout simulation) to reconstruct the entity via its constructor instead of writing to the now-read-only properties.

No behavior, validation logic, domain events, or public method signatures changed — this is a pure encapsulation refactor.

## Test plan

- [x] `pytest tests/unit/ -q` — 1973 passed, 0 failed
- [x] `ruff check .` — all checks passed
- [x] `mypy src/ --ignore-missing-imports` — 0 errors (baseline was already 0; verified via `git stash` that the pre-refactor tree also reports 0 errors, so no regression)
- [ ] Integration tests against a real Postgres DB were not run in this environment (no local Docker daemon available) — the mapper logic was verified by reading SQLAlchemy's `descriptor_props.py` to confirm `composite()` auto-derives `__composite_values__` from dataclasses (`Email`/`Password` are `@dataclass(frozen=True)`), and by mirroring the exact, already-tested `Competition` mapper pattern byte-for-byte in structure.

Closes #109

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of competition and user searches, including creator and name filtering.
  * Fixed verification-token updates to use the user’s domain behavior, ensuring consistent persistence.
  * Preserved expected password reset, email verification, account lockout, and profile update behavior.

* **Refactor**
  * Strengthened encapsulation for user and password-history data while maintaining existing functionality.
  * Improved persistence mapping for user-related information and value objects.

### DIFF
--- a/src/modules/competition/infrastructure/persistence/sqlalchemy/competition_repository.py
+++ b/src/modules/competition/infrastructure/persistence/sqlalchemy/competition_repository.py
@@ -321,7 +321,7 @@ class SQLAlchemyCompetitionRepository(CompetitionRepositoryInterface):
 
         # Si hay búsqueda por creador, necesitamos hacer JOIN con User
         if search_creator:
-            query = query.join(User, Competition._creator_id == User.id)
+            query = query.join(User, Competition._creator_id == User._id)
 
         # Lista de condiciones WHERE
         conditions = []
@@ -334,8 +334,8 @@ class SQLAlchemyCompetitionRepository(CompetitionRepositoryInterface):
         if search_creator:
             conditions.append(
                 or_(
-                    User.first_name.ilike(f"%{search_creator}%"),
-                    User.last_name.ilike(f"%{search_creator}%"),
+                    User._first_name.ilike(f"%{search_creator}%"),
+                    User._last_name.ilike(f"%{search_creator}%"),
                 )
             )
 
--- a/src/modules/user/application/use_cases/resend_verification_email_use_case.py
+++ b/src/modules/user/application/use_cases/resend_verification_email_use_case.py
@@ -105,7 +105,7 @@ class ResendVerificationEmailUseCase:
                 raise ResendVerificationError(GENERIC_ERROR_MESSAGE)
 
             # Asignar el token generado fuera de transacción
-            user.verification_token = verification_token
+            user.set_verification_token(verification_token)
 
             await self._uow.users.save(user)
             # El context manager (__aexit__) hace commit automático
--- a/src/modules/user/domain/entities/password_history.py
+++ b/src/modules/user/domain/entities/password_history.py
@@ -51,11 +51,31 @@ class PasswordHistory:
             created_at: Timestamp de creación (None para NOW())
             domain_events: Eventos de dominio pendientes
         """
-        self.id = id or PasswordHistoryId.generate()
-        self.user_id = user_id
-        self.password_hash = password_hash
-        self.created_at = created_at or datetime.now()
+        self._id = id or PasswordHistoryId.generate()
+        self._user_id = user_id
+        self._password_hash = password_hash
+        self._created_at = created_at or datetime.now()
         self._domain_events = domain_events or []
+
+    # ===========================================
+    # PROPERTIES (Encapsulación — solo lectura)
+    # ===========================================
+
+    @property
+    def id(self) -> PasswordHistoryId:
+        return self._id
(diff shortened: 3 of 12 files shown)


agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai full review

PR #138, opened 2026-07-30
refactor(social): Translate SQLAlchemy IntegrityError into a domain exception in BlockUserUseCase

### LINKED ISSUE #134: [Refactor]: Translate SQLAlchemy IntegrityError into a domain exception in BlockUserUseCase
## What part of the code is being refactored?

`src/modules/social/application/use_cases/block_user_use_case.py:3,57-64` imports `sqlalchemy.exc.IntegrityError` and catches it directly to detect a concurrent-block race on the `uq_friendship_pair` unique index, then calls `self._uow.rollback()` and re-queries.

## Motivation

Found during a follow-up Clean Architecture/DDD review of the features merged since the original backend audit (#106): friends system (#115), quick match module (#116, #117, #119, #121), and avatar (#125).

This is the only occurrence of `IntegrityError` anywhere in `src/` (confirmed via `grep -rn "IntegrityError" src/`) — every other use case in the codebase depends only on the UoW/repository *interfaces*, never on a concrete ORM's exception hierarchy. This breaks the Port/Adapter boundary the Unit-of-Work pattern exists to provide.

Concretely: the in-memory `InMemorySocialUnitOfWork` (used in unit tests) has no equivalent unique-constraint check and will never raise `IntegrityError`, so this race-handling branch has zero unit-test coverage and is only exercised against real Postgres — a latent bug here would only surface in integration/production.

Proposed fix: translate the constraint violation inside the SQLAlchemy repository/UoW adapter into a domain-level exception (e.g. `DuplicateFriendshipViolation`, following the naming pattern already used in `social/domain/exceptions/social_violations.py`), and have `InMemorySocialUnitOfWork` raise the same domain exception on the equivalent race so the branch becomes testable without a real DB.

## Risk / impact on existing tests

Low-medium — behavior for the end user is unchanged (still a rollback + re-query on race), but the exception translation touches both the SQLAlchemy adapter and the in-memory test double, and should add the first unit test that actually exercises this race-handling branch.

## Priority

P2 — Normal.

## Effort estimate

**S (~half a day).** Add `DuplicateFriendshipViolation` to `social/domain/exceptions/social_violations.py`, catch+translate `IntegrityError` inside the SQLAlchemy friendship repository, add the equivalent race-detection + raise to `InMemorySocialUnitOfWork`, update `BlockUserUseCase` to catch the domain exception instead, and add a unit test covering the race path (impossible today since the in-memory double can't trigger it).

agustinEDev:
## Summary
- `BlockUserUseCase` imported and caught `sqlalchemy.exc.IntegrityError` directly to handle a concurrent-block race on `uq_friendship_pair`, breaking the Port/Adapter boundary the Unit-of-Work exists to provide. It was also the only occurrence of `IntegrityError` in `src/`.
- Added `DuplicateFriendshipViolation` domain exception (`social_violations.py`).
- `SQLAlchemySocialUnitOfWork.flush()` now catches `IntegrityError`, rolls back, and translates it to `DuplicateFriendshipViolation` when the `uq_friendship_pair` constraint is involved (other integrity errors are re-raised unchanged).
- `InMemoryFriendshipRepository.add()` now enforces the same "one relationship per pair, any status/direction" invariant as the real unique index, raising the same domain exception — matching the DB adapter's behavior for tests.
- `BlockUserUseCase` now catches `DuplicateFriendshipViolation` instead of the SQLAlchemy exception.
- Rewrote `test_block_handles_concurrent_insert_race_idempotently` so the race is exercised through the real (now duplicate-checking) in-memory `add()`, only patching `find_by_pair` to simulate the read-before-concurrent-commit race window — previously the test had to directly force-raise `IntegrityError` via a monkeypatched `add()`, since the in-memory double had no way to produce it.

Closes #134

## Test plan
- [x] `pytest tests/unit/modules/social/` — 77 passed
- [x] `ruff check` / `ruff format` — clean
- [x] `mypy` — no issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of simultaneous attempts to block the same person.
  * Prevented duplicate friendship records in both in-memory and database-backed storage.
  * Preserved idempotent behavior when a block request races with another request.

### DIFF
--- a/src/modules/social/application/use_cases/block_user_use_case.py
+++ b/src/modules/social/application/use_cases/block_user_use_case.py
@@ -1,11 +1,12 @@
 """Caso de Uso: Bloquear a un usuario."""
 
-from sqlalchemy.exc import IntegrityError
-
 from src.modules.social.application.dto.friendship_dto import FriendshipResponseDTO
 from src.modules.social.application.exceptions import AddresseeNotFoundError
 from src.modules.social.domain.entities.friendship import Friendship
-from src.modules.social.domain.exceptions.social_violations import SelfFriendRequestViolation
+from src.modules.social.domain.exceptions.social_violations import (
+    DuplicateFriendshipViolation,
+    SelfFriendRequestViolation,
+)
 from src.modules.social.domain.repositories.social_unit_of_work_interface import (
     SocialUnitOfWorkInterface,
 )
@@ -57,10 +58,10 @@ class BlockUserUseCase:
                 try:
                     await self._uow.friendships.add(friendship)
                     await self._uow.flush()
-                except IntegrityError:
+                except DuplicateFriendshipViolation:
                     # Concurrent block on the same pair: another request already
-                    # won the uq_friendship_pair race. Idempotent: reload and return it.
-                    await self._uow.rollback()
+                    # won the uq_friendship_pair race (translated by the UoW
+                    # adapter). Idempotent: reload and return it.
                     friendship = await self._uow.friendships.find_by_pair(blocker_id, blocked_id)
             elif not friendship.is_blocked():
                 friendship.block(blocker_id)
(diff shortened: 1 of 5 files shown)


agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai review

PR #150, opened 2026-08-02
feat(admin): Add last login to admin user list

agustinEDev:
## Summary
- Adds `last_login_at` to the admin panel user list, reusing the existing device-fingerprinting `last_used_at` tracking instead of a new column/migration.
- New `UserDeviceRepositoryInterface.find_last_login_map(user_ids)` aggregates `MAX(last_used_at)` per user (SQLAlchemy + in-memory implementations).
- `AdminListUsersUseCase` attaches it to `AdminUserSummaryDTO.last_login_at` (nullable — users without any tracked device get `None`).

Supports the Competitions tab admin panel work on the frontend (RyderCupWeb #282), which bundles this field into the same Users tab.

## Test plan
- [x] `pytest tests/unit/ -q -n auto` — 2321 passed
- [x] `ruff check` on touched files — clean
- [x] `mypy --ignore-missing-imports` on touched files — clean
- [ ] CI green (unit + integration + lint + type + Snyk)

## Summary by CodeRabbit

* **New Features**
  * Admin user listings now display each user’s most recent login time.
  * Users without registered devices show no login timestamp.
  * Login activity is calculated across both active and inactive devices.

* **Tests**
  * Added coverage for latest login selection and users without device activity.

### DIFF
--- a/src/modules/user/application/dto/admin_dto.py
+++ b/src/modules/user/application/dto/admin_dto.py
@@ -37,6 +37,7 @@ class AdminUserSummaryDTO(BaseModel):
     is_active: bool
     email_verified: bool
     created_at: datetime
+    last_login_at: datetime | None = None
 
 
 class AdminListUsersResponseDTO(BaseModel):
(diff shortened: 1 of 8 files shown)


agustinEDev:
Added a follow-up commit: `PUT /admin/users/{id}` (`AdminUpdateUserUseCase`) reused `AdminUserSummaryDTO` but never populated `last_login_at`, so every admin edit silently wiped it back to `null" in the response. Flagged by CodeRabbit on RyderCupWeb#283 (the field's only current consumer) — fixed here at the source.

Issue #127, opened 2026-07-30
[Feature]: Allow a participant to hide a Quick Match from their own history


agustinEDev:
## Description
Originally scoped as a physical `DELETE /quick-matches/{id}` endpoint. Superseded (2 Aug 2026) by a softer per-participant design: a shared record shouldn't be deleted just because one participant wants it out of their own history — that would remove it for every other participant too. Instead, each participant can hide the match from their own `/quick-matches` list without touching the record, its scores, or anyone else's view.

- New field on `QuickMatch`: `hidden_by_participant_ids` (JSONB list of `ParticipantId`) — participants who've hidden this match from their personal history.
- Any participant (not just the creator) can hide/unhide the match for themselves — it's a personal preference, not an ownership action.
- No cascade delete: `QuickMatchHoleScore` rows and the match itself stay intact for every other participant, and for future aggregated stats (see #128).

## Acceptance criteria
- [ ] `hidden_by_participant_ids` field added to `QuickMatch` entity + persistence (migration)
- [ ] New use case (e.g. `HideQuickMatchUseCase`) — any participant can hide/unhide, only affecting their own view
- [ ] `GET /quick-matches/me` excludes matches where the calling participant's `participant_id` is in `hidden_by_participant_ids`
- [ ] Unit + integration tests (authorization: any participant; only affects the caller; idempotent hide/unhide)

## Priority
P2 — Normal.

## Additional context
Paired with RyderCupWeb#263 (trash-can button now calls hide instead of delete). Design decision made 2 Aug 2026.

PR #169, opened 2026-08-09
feat(api): Player statistics and recent matches endpoints

### LINKED ISSUE #128: feat(api): Player statistics and recent matches endpoints (dashboard + per golf course)
> Unifica esta issue con la #163, abierta a raíz de la auditoría externa del dashboard móvil (agosto 2026). Ambas definían `GET /api/v1/users/me/stats`. #163 queda cerrada como duplicada.

## Descripción

No existe todavía ningún concepto de estadísticas de jugador en el backend (el único "stats" actual es `ScorecardStatsDTO`, por partido, no agregado sobre el historial). Hay que diseñar y añadir los endpoints que expongan el rendimiento agregado de un jugador, tanto para el resumen del dashboard como para el desglose por campo.

Consumidor principal: **RyderCupWeb #306** (rediseño del dashboard móvil).

## Diseño previo (necesario antes de implementar)

- Qué fuentes alimentan la agregación: `Match`/`HoleScore` (módulo competition), `QuickMatch`/`QuickMatchHoleScore` (módulo quick_match), o ambas.
- Métricas candidatas: rondas/partidos jugados, media respecto al par, media de puntos Stableford, mejor ronda, tendencia de hándicap, medias por campo (agrupando por `golf_course_id`).
- Si merece un módulo de read-model compartido (agregación entre módulos) o extender el módulo `user`, dado que lee de `competition` y `quick_match`.
- **Partidas ocultas (#127):** una partida rápida que el jugador ha ocultado de su historial (`hidden_by_participant_ids`) debe quedar excluida también de SUS estadísticas agregadas. El ocultado es por participante, así que **no** puede afectar a las estadísticas de los demás participantes de esa misma partida.

## Endpoints

**`GET /api/v1/users/me/stats`** — resumen global
```
{ handicap, handicapTrend, scoringAvg, roundsPlayed, tournamentsTotal, tournamentsActive }
```

**`GET /api/v1/users/me/matches?limit=3`** — últimas partidas para el feed del dashboard
```
[{ id, date, format, course, tournamentName?, result, score, partners[], opponents[] }]
```

**`GET /api/v1/users/me/stats/golf-courses/{golf_course_id}`** (o equivalente) — desglose por campo, con la misma regla de exclusión.

## El modelo hay que rediseñarlo, no copiarlo

El contrato que traía la auditoría usa `format: "match_play" | "stroke_play" | "fourball"`, mezclando dos ejes que aquí son distintos y mutuamente excluyentes:

- `MatchFormat`: SINGLES / FOURBALL / FOURSOMES
- `ScoringFormat`: MEDAL / STABLEFORD

Hay que decidir cómo se representa esto en el DTO, y cómo se unifican partidas rápidas y partidos de competición en un mismo feed. La propuesta original usaba además `/api/users/me/...` sin el `/v1`: se respeta la convención del proyecto.

## Lo más caro: `handicapTrend`

Iba disimulado en una línea del documento externo, pero **es el punto más costoso**: requiere histórico de hándicap, que hoy no existe (solo `handicap_updated_at`, sin snapshots). Implica tabla nueva, migración y escritura en cada actualización.

Sugerencia: dejarlo fuera de la primera entrega devolviendo `handicapTrend: null` — el frontend ya contempla ese caso.

## Criterios de aceptación

- [ ] Documento de diseño/ADR acordado antes de picar código: fuentes, métricas, estrategia de agregación (consulta al vuelo vs. precomputado)
- [ ] `GET /users/me/stats` — global, excluyendo las partidas ocultas por quien llama (#127)
- [ ] `GET /users/me/matches?limit=N` — feed de últimas partidas, misma exclusión
- [ ] `GET /users/me/stats/golf-courses/{id}` — desglose por campo, misma exclusión
- [ ] Datos reales para una cuenta con historial; `null`/listas vacías sin romper para cuentas nuevas
- [ ] Tests unitarios + integración (incluido: una partida oculta no cuenta para quien la ocultó, pero sí para el resto)

## Prioridad

P1 — sube de P2 porque bloquea el rediseño del dashboard (RyderCupWeb #306).

agustinEDev:
Closes #128

Adds the aggregated player performance the dashboard redesign needs (RyderCupWeb #306). Everything is computed on the fly — a player's history is a handful of rounds, it changes with every score recorded, and an invalidation strategy would cost more than the query it saves.

## Endpoints

- `GET /api/v1/users/me/stats` — global summary
- `GET /api/v1/users/me/stats/golf-courses/{id}` — same summary, one course only
- `GET /api/v1/users/me/matches?limit=N` — recent match feed

Responses stay snake_case like the rest of the API; the camelCase contract in the issue came from an external audit document, not from this project. An account with no history returns zeros, nulls and an empty list rather than a 404: an empty dashboard is a normal state, not an error.

## What reaches the statistics

**Only complete scorecards from finished matches.** A match still in progress, a card left half scored or a single unscored hole keeps the round out of `scoring_avg` *and* of `rounds_played` — the two numbers must always talk about the same rounds, and half a round cannot be compared with a full one. Completeness is measured against the holes the course actually has, not a hardcoded 18.

What decides is the scorecard, not the hole the match was won on: a match play resolved on the 15th counts like any other round as long as the players kept scoring to the 18th. The statistics never look at `is_decided`. What leaves a round out is giving up on the card, which in match play is the common case.

The feed (`/me/matches`) does not apply this filter — it is history, not a statistic, so a finished match with a partial card still shows up there.

## Design decisions

**Stableford moved into the backend.** The scoring rules only lived in the frontend. `StablefordCalculator` ports them, reusing `PlayingHandicapCalculator` — which gained `calculate_unbounded()`, without the `max(0, ...)`, so a plus player gives strokes as WHS Rule 8.2 says. Parity with the frontend engine is pinned by tests over the same data across five handicaps, one of them a plus.

**Both sources feed the average.** A tournament match is played hole by hole against a rival, but the player still signs a card with their own strokes, and that card says how well they played just as much as a medal round does — so Stableford tournaments will work for free when we add them. It reads `own_score`, not the entity's `net_score`: the latter is only filled in once the marker has validated the hole, so good cards would silently drop out of the average.

**Each hole is capped at net double bogey** (WHS Rule 3.1), on both sources so they stay comparable — one blow-up hole should not move a whole season's average.

**The feed unifies two different things.** Each entry leaves as null whatever does not apply: a tournament match has no `scoring_format`, a free-play match has no match-play `result`. `match_format` and `scoring_format` are not merged into a single `format` field because in the domain they are distinct, mutually exclusive axes.

**Use cases live in the `user` module**, taking the units of work of the modules they read, following the existing `GetAdminStatsUseCase` pattern.

## Out of scope

`handicap_trend` returns `null`: there is no handicap history, only the date of the last change. Follow-ups opened from this work: #167 (score differentials and an estimated index, which also gets the trend without a new table) and #168 (breakdown by par, hole type and nines).

## Tests

2425 unit tests plus 15 integration tests. `find_completed_for_player` is verified against PostgreSQL, since it looks the player up with JSONB containment over two columns with no foreign key and orders by the round date from another table — the in-memory implementation cannot show that. The endpoint tests play a full quick match through the API and then read the three endpoints, including the #127 rule: hiding a match removes it from the stats and the feed of whoever hid it, and only for them.

### DIFF
--- a/src/config/dependencies.py
+++ b/src/config/dependencies.py
@@ -318,6 +318,12 @@ from src.modules.user.application.use_cases.get_current_user_use_case import (
 from src.modules.user.application.use_cases.get_my_avatar_upload_image_use_case import (
     GetMyAvatarUploadImageUseCase,
 )
+from src.modules.user.application.use_cases.get_player_stats_use_case import (
+    GetPlayerStatsUseCase,
+)
+from src.modules.user.application.use_cases.get_recent_matches_use_case import (
+    GetRecentMatchesUseCase,
+)
 from src.modules.user.application.use_cases.google_login_use_case import (
     GoogleLoginUseCase,
 )
@@ -1200,6 +1206,26 @@ def get_get_admin_stats_use_case(
 ) -> GetAdminStatsUseCase:
     """Proveedor del caso de uso GetAdminStatsUseCase."""
     return GetAdminStatsUseCase(user_uow, competition_uow, quick_match_uow, golf_course_uow)
+
+
+def get_get_player_stats_use_case(
+    user_uow: UserUnitOfWorkInterface = Depends(get_uow),
+    competition_uow: CompetitionUnitOfWorkInterface = Depends(get_competition_uow),
+    quick_match_uow: QuickMatchUnitOfWorkInterface = Depends(get_quick_match_uow),
+    golf_course_uow: GolfCourseUnitOfWorkInterface = Depends(get_golf_course_uow),
+) -> GetPlayerStatsUseCase:
+    """Proveedor del caso de uso GetPlayerStatsUseCase."""
+    return GetPlayerStatsUseCase(user_uow, competition_uow, quick_match_uow, golf_course_uow)
+
+
+def get_get_recent_matches_use_case(
+    user_uow: UserUnitOfWorkInterface = Depends(get_uow),
+    competition_uow: CompetitionUnitOfWorkInterface = Depends(get_competition_uow),
+    quick_match_uow: QuickMatchUnitOfWorkInterface = Depends(get_quick_match_uow),
+    golf_course_uow: GolfCourseUnitOfWorkInterface = Depends(get_golf_course_uow),
+) -> GetRecentMatchesUseCase:
+    """Proveedor del caso de uso GetRecentMatchesUseCase."""
+    return GetRecentMatchesUseCase(user_uow, competition_uow, quick_match_uow, golf_course_uow)
 
 
 def get_create_quick_match_use_case(
--- a/src/modules/competition/domain/repositories/match_repository_interface.py
+++ b/src/modules/competition/domain/repositories/match_repository_interface.py
@@ -2,6 +2,8 @@
 
 from abc import ABC, abstractmethod
 
+from src.modules.user.domain.value_objects.user_id import UserId
+
 from ..entities.match import Match
 from ..value_objects.match_id import MatchId
 from ..value_objects.round_id import RoundId
@@ -23,6 +25,20 @@ class MatchRepositoryInterface(ABC):
     @abstractmethod
     async def find_by_id(self, match_id: MatchId) -> Match | None:
         """Busca un partido por su ID."""
+        pass
+
+    @abstractmethod
+    @abstractmethod
+    async def find_completed_for_player(
+        self, user_id: UserId, limit: int | None = None
+    ) -> list[Match]:
+        """
+        Partidos terminados en los que el usuario jugó, del más reciente al más
+        antiguo por fecha de la ronda.
+
+        Alimenta las estadísticas e historial del jugador (BE #128). Solo
+        cuenta lo terminado: un partido a medias todavía no dice cómo quedó.
+        """
         pass
 
     @abstractmethod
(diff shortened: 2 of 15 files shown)


agustinEDev:
Thanks for the review. Addressed in 96cb0de:

**1. `MAX_ROUNDS_AGGREGATED` on the competition query — fixed.** Real asymmetry: the quick match side was capped and the competition side was not, while the competition side is the more expensive of the two (a round lookup plus a scorecard query per match). Applied as suggested; the cap is per source, so the aggregation is bounded at 200 rounds.

**2. Hard-coded 18 holes — documented, not derived.** `GolfCourse._validate_holes` rejects any course that does not have exactly 18 holes (`golf_course.py:507`), so the constant and the course are the same number by construction — deriving it would add indirection without changing a single result. Took the option you offered: both constants now state the invariant and what to change if nine-hole courses are ever admitted. Note that `_is_full_scorecard` in the stats use case *is* course-driven, deliberately, because there the cost of being wrong is silently dropping a legitimate round.

**3. In-memory repository ordering — not changing it.** `InMemoryMatchRepository` is a test double with no rounds to ask; giving it round dates would mean wiring a second in-memory repository into it. The ordering contract is verified against the real SQLAlchemy implementation by `test_orders_by_round_date_most_recent_first` and `test_limit_keeps_the_most_recent_ones`, and the divergence is already stated in the method's docstring.

PR #170, opened 2026-08-09
feat(stats): Score differential and estimated handicap index

### LINKED ISSUE #167: feat(stats): Score differential and estimated handicap index from played rounds
> Sale de la conversación de BE #128: además de la media respecto al par, poder decirle al jugador **a qué hándicap está jugando** sus últimas rondas.

## Idea

El WHS ya tiene la métrica exacta, el **Score Differential**, que es literalmente el hándicap al que jugaste esa vuelta:

```
Differential = (113 / Slope) × (Adjusted Gross Score − Course Rating)
```

Con los últimos 20 diferenciales, la media de los **8 mejores** es un Handicap Index. Enseñado junto al hándicap oficial del perfil da el titular del dashboard: *"tu hándicap es 14,2 — estás jugando a 12,8"*.

Eso resuelve además el `handicap_trend` que la primera entrega dejó en `null`: la tendencia sale de los diferenciales de las rondas, **sin necesidad de tabla de histórico de hándicap**.

## Los datos ya están (comprobado)

- **Golpes por hoyo:** `QuickMatchHoleScore.score` y `HoleScore.own_score` + `strokes_received`.
- **Course Rating / Slope:** en `Tee`. El tee jugado está en `MatchPlayer.tee_category`/`tee_gender` (siempre, en torneo) y en `QuickMatchParticipant.tee_category`/`tee_gender` (opcional en el dominio, pero el modal de creación del frontend lo exige desde el 28 jul; las partidas anteriores a esa fecha pueden no tenerlo).
- **Par por hoyo:** `Hole.par`.
- **Adjusted Gross Score:** ya existe `StablefordCalculator.adjusted_gross()`, que aplica el tope de net double bogey (Regla WHS 3.1).

## Alcance propuesto

- [ ] `ScoreDifferentialCalculator` en el dominio, puro y sin IO
- [ ] Solo entran vueltas de **18 hoyos** anotados. Las de 9-17 no computan para el diferencial (el escalado a 18 del WHS es otra historia); las parciales siguen contando para el resto de métricas
- [ ] Una vuelta **sin tee conocido no genera diferencial**: no hay Slope ni Course Rating con los que calcularlo. Hay que decidir qué se enseña en ese caso
- [ ] `estimated_index`: media de los 8 mejores de los últimos 20 diferenciales; `null` con menos de 3 rondas (el WHS necesita 54 hoyos, y con dos vueltas el número sería ruido)
- [ ] `playing_avg`: media de los últimos N diferenciales, sin quedarse con los mejores
- [ ] `best_differential` / mejor vuelta
- [ ] `handicap_trend`: comparar los últimos diferenciales con los anteriores
- [ ] Serie de los últimos N diferenciales, para que el dashboard pueda pintar una línea

## Ojo al presentarlo

No hay **PCC** (Playing Conditions Calculation), así que se asume 0 y el índice **no es el oficial de la RFEG**, es el cálculo de la app. Conviene que la UI lo diga para no generar la expectativa equivocada.

## Prioridad

P2 — el dashboard (RyderCupWeb #306) ya se puede montar con lo entregado en BE #128; esto es lo que lo hace interesante.

agustinEDev:
Closes #167.

Tells a player **what handicap they are actually playing to**, from the rounds they have already played. `handicap_trend`, which BE #128 had to leave `null`, now has an answer — and it needed no handicap history table after all.

## What the API gains

`GET /api/v1/users/me/stats` (and its per-course variant) return five new fields:

| Field | What it says |
|---|---|
| `estimated_index` | The handicap they are playing to. `null` under 3 rounds |
| `playing_avg` | Average of all recent differentials, not just the best ones |
| `best_differential` | Their best round |
| `rounds_with_differential` | How many rounds the index could actually use |
| `differentials` | The series, newest first, for the dashboard to plot |

And `handicap_trend` is finally filled in.

## Decisions worth reviewing

**The Rule 5.2 table, not a flat "best 8 of 20".** The issue described the 20-round case. Applied literally with five rounds it means averaging all of them, which reads several strokes better than the player is. The WHS table takes only the best one and subtracts a margin while the sample is thin, so that is what this implements.

**The differential computes its own Adjusted Gross Score.** It does not reuse the one behind `scoring_avg`. Two caps, on purpose: the average is our own metric and uses the handicap the match was played off, allowance included; the differential means to be WHS, and the WHS ignores the allowance because it measures a round against the course, not against the rivals in it.

**A round from an unknown tee yields no differential.** No Slope, no Course Rating, nothing to compute. It still counts towards the average, so `rounds_with_differential` can be lower than `rounds_played` — otherwise the index would look like it covers every round a player has played. Quick matches created before the frontend started requiring a tee land here.

**`handicap_trend` is negative when the player improves.** It is the raw difference between the last five differentials and the five before them. Differentials fall as you play better, exactly like a handicap falls, so inverting the sign would make the number contradict the handicap shown next to it.

**Tournament rounds use `MatchPlayer.player_handicap`**, the snapshot taken when the match was generated, so a round from months ago is not judged by today's handicap. Quick matches have no such snapshot and fall back to the profile.

## Not the official index

There is no PCC — the playing conditions adjustment needs every card handed in that day. Assumed 0, so this is the app's calculation, not the federation's. The DTO says so, and the UI should too.

## Verification

2473 unit tests, 366 integration, 45 security — all green locally against the Kind cluster database. `ruff` and `mypy` clean.

The expected numbers in the tests were worked out by hand from the WHS formula before running anything, not read back off the implementation: `(113/125) x (90 - 70.0) = 18.1`, `(113/130) x (90 - 72.0) = 15.6`, and a round of eights capped at net double bogey giving `42.5` instead of the `66.9` it would be uncapped.

## Summary by CodeRabbit

* **New Features**
  * Added WHS score differential calculations, estimated indexes, playing averages, best differentials, and performance trends.
  * Player statistics now include differential history, differential count, estimated index, playing average, and best differential.
  * Differentials are calculated for completed rounds with valid tee information.
  * Adjusted gross scores are now included in participant scoring totals.

* **Bug Fixes**
  * Improved handling of rounds without tee data and insufficient rounds for handicap calculations.

### DIFF
--- a/src/modules/competition/domain/services/score_differential_calculator.py
+++ b/src/modules/competition/domain/services/score_differential_calculator.py
@@ -1,0 +1,158 @@
+"""
+Domain Service: ScoreDifferentialCalculator.
+
+Score Differential del WHS: el hándicap al que un jugador jugó una vuelta
+concreta, y el índice estimado que sale de sus últimas vueltas. Puro, sin IO.
+
+Vive junto a `PlayingHandicapCalculator` porque son la misma familia de reglas
+WHS y comparten `TeeRating`, aunque ninguno de los dos sea exclusivo del módulo
+de competición.
+
+**Este índice no es el oficial de la federación.** Falta el PCC (Playing
+Conditions Calculation), que ajusta los diferenciales por cómo estaba el campo
+ese día y que solo puede calcular quien tiene todas las tarjetas de la jornada.
+Aquí se asume PCC = 0.
+"""
+
+from dataclasses import dataclass
+from decimal import ROUND_HALF_UP, Decimal
+
+from src.modules.competition.domain.services.playing_handicap_calculator import (
+    NEUTRAL_SLOPE,
+    TeeRating,
+)
+
+# Ventana de vueltas del WHS: el índice mira las 20 últimas, no la carrera
+# entera. Un jugador que mejora no arrastra sus vueltas de hace tres años.
+SCORING_RECORD_SIZE = 20
+
+# Sin al menos estas vueltas no se publica índice. El WHS pide 54 hoyos, que
+# son tres vueltas: con dos, el número es ruido con aspecto de dato.
+MIN_ROUNDS_FOR_INDEX = 3
+
+# Cuántas vueltas se promedian y qué ajuste se les aplica, según cuántas hay
+# disponibles (Regla WHS 5.2). Con pocas vueltas se coge solo la mejor y se le
+# resta un margen: la muestra es tan pequeña que sin ese ajuste el índice
+# saldría optimista. Cada entrada es (mínimo de vueltas, cuántas se promedian,
+# ajuste), ordenada de más a menos vueltas.
+_INDEX_TABLE: tuple[tuple[int, int, Decimal], ...] = (
+    (20, 8, Decimal("0")),
+    (19, 7, Decimal("0")),
+    (17, 6, Decimal("0")),
+    (15, 5, Decimal("0")),
+    (12, 4, Decimal("0")),
+    (9, 3, Decimal("0")),
+    (7, 2, Decimal("0")),
+    (6, 2, Decimal("-1.0")),
+    (5, 1, Decimal("0")),
+    (4, 1, Decimal("-1.0")),
+    (3, 1, Decimal("-2.0")),
+)
+
+# Cuántas vueltas entran en cada mitad de la comparación de tendencia. Cinco es
+# suficiente para que una vuelta mala no dicte la tendencia, y bastante poco
+# para que la tendencia hable del presente.
+TREND_WINDOW = 5
+
+
+@dataclass(frozen=True)
+class PlayedRound:
+    """
+    Una vuelta ya reducida a lo que el WHS necesita de ella.
+
+    `adjusted_gross_score` son los golpes de la vuelta con cada hoyo topado en
+    su net double bogey; el tope se aplica antes, al leer la tarjeta, porque
+    depende de los golpes que el jugador recibía en cada hoyo.
+    """
+
+    adjusted_gross_score: int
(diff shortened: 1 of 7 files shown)


agustinEDev:
Thanks for the review. Addressed in 9d7e540:

**1. `differentials` vs the 20-round window — fixed, and this was the one that mattered.** A real inconsistency observable from outside: `min(differentials)` in a client could return a lower number than `best_differential`, and a chart would plot rounds no other figure was counting. Truncating rather than documenting, because the series exists to draw the trend and the trend is exactly what the other metrics measure. The window now lives in one place, `ScoreDifferentialCalculator.scoring_record`, which every metric routes through, so the two cannot drift apart again. `rounds_with_differential` deliberately still counts every round that has one — that is what makes it worth publishing next to `rounds_played`. Pinned by a test with 21 rounds asserting `min(differentials) == best_differential`.

**2. Soft cap and hard cap — documented.** Correct that the docstring was misleading: it named PCC as the only omission. Not implementing them, because the Low Handicap Index means reconstructing what the index was at every point of the last 365 days to find its minimum, which is a good deal more than a formula and worth its own issue if it is ever wanted. Worth noting the direction of the error: without the caps a player who gets worse sees a *higher* index here than the official system would give, so the number never flatters. That is now in the docstring too.

**3. Re-entering `golf_course_uow` — not changing it.** `SQLAlchemyGolfCourseUnitOfWork.__aenter__` returns `self` and starts no transaction, so the two sequential blocks are safe as the code stands. The concern is conditional on a future change to the session handling, and the pattern is not introduced here — it has been in `develop` since #128, and `GetRecentMatchesUseCase` does the same. If the session handling ever becomes transaction-scoped, this needs fixing everywhere at once, not in one read-only use case.

Issue #173, opened 2026-08-10
fix(stats): Incomplete cards bias every statistic optimistic


agustinEDev:
Found while diagnosing a real production account (10 Aug 2026).

## The effect

Statistics only count **whole rounds of finished matches** — the rule set in #128, and the right one for the average: half a card cannot be compared with a full one.

But looking at real data, that filter has a bias I had not anticipated: **the rounds people abandon are the bad ones**. Nobody stops scoring when they are playing well.

The account that surfaced it had four matches. The worst — 20 Stableford points, when 36 is playing to your handicap — is the one with the incomplete card, so it does not appear in `rounds_played`, in `scoring_avg`, or in any differential. Every figure the player sees is computed from their better days.

## Why it matters

It is not a calculation error, it is a sampling one, and it compounds:

- `scoring_avg` reads better than the player actually scores.
- `estimated_index` reads better still, because on top of the biased sample it takes only the best rounds.
- The player sees numbers that do not match their own sense of how they are playing — which is exactly the report that led here.

## Not obvious what the fix is

Options worth weighing, rather than picking one now:

1. **Encourage finishing the card.** Already noted as a product idea when #167 landed: prompt to keep scoring once a match is decided ("three holes left for this round to count"). Attacks the cause rather than the symptom, and needs frontend work.
2. **Let a round count with holes missing**, filling them by some rule. The WHS has a mechanism for this, but it is not free: it changes what `scoring_avg` means and needs care not to reward abandoning.
3. **Say it in the interface**: show how many rounds were left out for being incomplete, the way `rounds_with_differential` already does for missing tees. Cheapest, and honest, but does not improve the numbers.

Option 1 plus option 3 look like the sensible pair, but this deserves a decision rather than an implementation straight away.

## Related

- #128 — where the whole-cards-only rule was set, and why
- #167 — the estimated index, which amplifies the bias
- RyderCupWeb#351 — the label fix that came from the same report

PR #178, opened 2026-08-10
feat(social): Activity events generated from finished rounds

### LINKED ISSUE #175: feat(social): Activity events generated from finished rounds
Primera pieza del componente social (ver RyderCupWeb#306 para el contexto del panel). Aquí solo el motor: **qué se publica y de dónde sale**. El feed y los perfiles van en issues aparte.

## La regla que gobierna el diseño

**Se publican logros, no actividad.** Un birdie apetece enseñarlo; "jugó 108 golpes el martes" no. Y el sistema que publica lo segundo empuja a la gente a no anotar sus vueltas malas — que es exactamente el sesgo que ya tenemos abierto en #173. Nada de lo que se publique debe desincentivar terminar una tarjeta.

## Eventos iniciales

Todos salen de datos que ya tenemos, sin capturar nada nuevo:

| Evento | De dónde sale |
|---|---|
| `HOLE_IN_ONE` | Un hoyo con 1 golpe |
| `EAGLE_OR_BETTER` | Golpes ≤ par − 2 |
| `BIRDIE` | Golpes = par − 1 |
| `NEW_COURSE` | Primer partido terminado en ese campo |
| `PERSONAL_BEST` | Mejor diferencial WHS del jugador hasta la fecha (#167) |
| `FIRST_TOURNAMENT` | Primera competición terminada |

Un birdie por vuelta es ruido si el jugador hace cuatro: **agrupar por vuelta**, un evento con su cuenta, no uno por hoyo.

## Cuándo se generan

Al **terminar** una partida (rápida o de torneo), sobre la misma tarjeta que ya alimenta las estadísticas. Solo cuentan las vueltas que cuentan para estadísticas (enteras o media vuelta limpia, ver #174): si una tarjeta no es fiable para la media, tampoco lo es para presumir.

## Modelo

- `ActivityEvent`: `id`, `user_id`, `type`, `occurred_at`, `payload` (JSONB con lo propio de cada tipo: campo, hoyo, cuántos birdies, diferencial…), `source_match_id`.
- Índice por `(user_id, occurred_at DESC)`: el feed siempre consulta así.
- **Idempotencia**: completar dos veces la misma partida no debe duplicar eventos. Clave única por `(source_match_id, type)`.

## Interruptor

`users.share_activity` (booleano). Con él apagado no se generan eventos para ese jugador. Propongo que venga **activado** por defecto —es una app entre amigos y el feed nace vacío si no— pero es decisión de producto y conviene fijarla aquí.

## A decidir

- **¿Backfill?** ¿Se generan eventos de las partidas ya jugadas, o el feed empieza desde hoy? Con backfill el feed nace con contenido, pero publica de golpe cosas de hace meses, lo cual es raro de leer.

## Relacionado

- #174 — qué cuenta como vuelta
- #167 — el diferencial, del que sale `PERSONAL_BEST`
- #173 — el sesgo de las tarjetas incompletas, que esto no debe empeorar

agustinEDev:
Closes #175

The engine behind the friends feed: what gets published, where it comes from, and how a player opts out. The feed itself and the profiles are #176.

## The rule that governs the design

**Achievements get published, not activity.** A birdie is worth showing; "shot 108 on Tuesday" is not. And a system that publishes the second one pushes people to stop entering their bad rounds — exactly the bias already open in #173. A test pins the exact list of event types so that adding one is a deliberate decision.

## No backfill

Decided on the issue: the feed starts from today. Publishing months of history at once reads as noise, and the first thing every player would see is their own past instead of what their friends just did. Consequence: publishing only ever hooks into "round finished", and no migration walks past matches.

## What is included

| | |
|---|---|
| Domain | `ActivityEvent`, `ActivityEventType` (6 types), `AchievementDetector` (pure, no IO) |
| Persistence | `activity_events` table, migration `d4e8f1a3b7c9`, SQLAlchemy + in-memory repositories |
| Publishing | On closing a quick match and on closing a tournament |
| Opt-out | `PUT /api/v1/social/activity-sharing` |

`users` gains `feed_last_seen_at` (where the "something new" badge will come from, in #176) and `share_activity`.

## Decisions worth reviewing

**`source_match_id` is NOT NULL.** The unique key on `(user_id, source_match_id, type)` is the only thing stopping a reprocessed round from filling the feed with repeats, and in Postgres a NULL never equals another NULL — so every event without a match would have slipped past it. Verified against a real database: two rows with NULL both went in. Every achievement is born from a finished round, so requiring it excludes no real case.

**Inserts use `ON CONFLICT DO NOTHING`** rather than checking first and writing after. Checking leaves a window between "it does not exist" and the INSERT, and two requests completing the same match at once — a phone retrying over a bad connection — fall straight into it.

**`PERSONAL_BEST` is measured across the close, not recomputed.** The best differential is captured before the round is completed and compared after; if it improved, the only new round is that one. The WHS is implemented once, in the statistics module (#167), and a second implementation just for the feed would have diverged from it.

**Reading and judging are separate phases.** Asking for a differential re-enters `quick_match` and `golf_course` through the statistics module, so nesting that call inside an open unit of work would commit and close the session the outer block was still using.

**Which rounds count is now one rule in one place.** `countable_holes` moves to `shared` because statistics (#174) and the feed need the same answer: a card too incomplete for a player's average is also too incomplete to brag about.

**Turning sharing off removes what was already published.** Whoever flips that switch wants their activity to stop being visible, not to stop growing. Turning it back on restores nothing — keeping the removed events around for that would mean storing exactly what the player asked to take down.

## Module boundaries

`quick_match` and `competition` each declare a port; `social` implements it in an adapter. Neither module imports the other. A failure to publish is logged and swallowed: the feed is incidental, the scorecard is not, and nothing here may stop a round from being marked as finished.

## Verification

- **2539** unit tests, **426** integration and security tests, 1 skipped
- `mypy` clean across 572 files; `ruff` clean on everything touched
- Migration applied and rolled back against a real PostgreSQL, on a throwaway database in the dev cluster so the shared one kept its Alembic revision. Single Alembic head after the merge.

## Note for whoever tests this

Because there is no backfill, nothing appears in `activity_events` until a **new** round is closed. Rounds already finished before this ships generate no events, by design.

## Summary by CodeRabbit

* **New Features**
  * Added a social activity feed for achievements from quick matches and tournaments.
  * Added achievements for birdies, eagles, hole-in-ones, new courses, personal bests, and first tournaments.
  * Added controls to enable or disable activity sharing; disabling removes previously shared activity.
  * Added feed tracking and reliable duplicate prevention.
* **Bug Fixes**
  * Improved round scoring to consistently support complete and valid nine-hole rounds.
* **Tests**
  * Added comprehensive coverage for achievement publishing, feed privacy, duplicate prevention, and activity storage.

### DIFF
--- a/alembic/versions/d4e8f1a3b7c9_add_activity_events_table.py
+++ b/alembic/versions/d4e8f1a3b7c9_add_activity_events_table.py
@@ -1,0 +1,77 @@
+"""add activity_events table
+
+Revision ID: d4e8f1a3b7c9
+Revises: c7d1e4f8a2b6
+Create Date: 2026-08-10
+
+Feed de actividad entre amigos (BE #175). Guarda los logros que se publican:
+birdies, eagles, hoyos en uno, campos estrenados, récords personales y el
+primer torneo.
+
+Dos detalles del diseño que la tabla refleja:
+
+- `payload` es JSONB porque cada tipo de evento lleva un detalle distinto
+  (cuántos birdies y en qué hoyos, qué diferencial batió al anterior). Añadir un
+  tipo nuevo no debería exigir una migración.
+- La restricción única sobre `(user_id, source_match_id, type)` es lo que impide
+  que reprocesar una partida llene el feed de entradas repetidas.
+"""
+
+import sqlalchemy as sa
+from alembic import op
+from sqlalchemy.dialects import postgresql
+
+# revision identifiers, used by Alembic.
+revision = "d4e8f1a3b7c9"
+down_revision = "c7d1e4f8a2b6"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.create_table(
+        "activity_events",
+        sa.Column("id", postgresql.UUID(as_uuid=True), nullable=False),
+        sa.Column("user_id", sa.CHAR(36), nullable=False),
+        sa.Column("type", sa.String(30), nullable=False),
+        sa.Column("occurred_at", sa.DateTime(), nullable=False),
+        sa.Column(
+            "payload",
+            postgresql.JSONB(astext_type=sa.Text()),
+            nullable=False,
+            server_default="{}",
+        ),
+        sa.Column("source_match_id", sa.String(36), nullable=True),
+        sa.PrimaryKeyConstraint("id"),
+        sa.ForeignKeyConstraint(["user_id"], ["users.id"], ondelete="CASCADE"),
+        sa.UniqueConstraint(
+            "user_id", "source_match_id", "type", name="uq_activity_events_match_type"
+        ),
+    )
+    # El feed siempre pregunta lo mismo: eventos de estos jugadores, del más
+    # reciente al más antiguo
+    op.create_index(
+        "ix_activity_events_user_occurred",
+        "activity_events",
+        ["user_id", "occurred_at"],
+    )
+
+    # Cuándo miró cada jugador su feed por última vez. De aquí sale el aviso de
+    # novedades: contar lo publicado después de esta fecha. Nulo significa que
+    # todavía no lo ha abierto nunca.
+    op.add_column("users", sa.Column("feed_last_seen_at", sa.DateTime(), nullable=True))
+    # Interruptor para no publicar. Activado por defecto: es una aplicación entre
+    # amigos y el feed nacería vacío de otro modo.
+    op.add_column(
+        "users",
+        sa.Column(
+            "share_activity", sa.Boolean(), nullable=False, server_default=sa.true()
+        ),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("users", "share_activity")
+    op.drop_column("users", "feed_last_seen_at")
+    op.drop_index("ix_activity_events_user_occurred", table_name="activity_events")
+    op.drop_table("activity_events")
--- a/alembic/versions/d4e8f1a3b7c9_add_activity_events_table.py
+++ b/alembic/versions/d4e8f1a3b7c9_add_activity_events_table.py
@@ -58,6 +58,14 @@ def upgrade() -> None:
         "activity_events",
         ["user_id", "occurred_at"],
     )
+    # La comprobación de «esta partida ya se publicó» filtra solo por partida.
+    # El índice de arriba empieza por `user_id`, así que no le sirve, y sin este
+    # la comprobación acabaría recorriendo la tabla entera según crece el feed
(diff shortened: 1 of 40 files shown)


agustinEDev:
Thanks — six of the ten were real and are fixed in `5c00a7b`. Two I'm leaving as they are, with reasons.

## Fixed

**The critical one was the best catch.** Both completion flows captured differentials inside their own transaction, and that call reaches `GetPlayerStatsUseCase`, which reopens the same session-scoped units of work. Worth noting this is the exact re-entrancy the publishing use case already guards against — it was fixed there and left standing in both callers, which is precisely the kind of thing that survives a self-review. Reads now happen inside the unit of work, it closes, the capture runs, and it reopens to complete and persist.

**The pagination cursor was wrong and my comment about it was wrong too.** Events from one round all share `occurred_at` — they come from the same `created_at` — so this was the normal case, not an edge one. Now compares the `(occurred_at, id)` pair. Added an integration test that pages through four events sharing a timestamp and asserts each appears exactly once.

**The mapper guard never fired.** `mapper_registry.mappers` holds `Mapper` objects, not classes, so the check was always true; I confirmed a second call raises `already has a primary mapper defined`. Now uses `inspect(..., raiseerr=False)`. Note the same broken pattern exists in `start_social_mappers` — left alone here since it predates this branch, worth a separate issue.

**Sorting before `indice == 0`**: agreed, and the consequence is worse than ordering. If the database returned the matches differently between runs, `FIRST_TOURNAMENT` would hang off a different match and the unique key would stop recognising the repeat.

**Index on `source_match_id`** added to both the migration and the mapper. **`countable_holes`** now rejects a course with no holes explicitly.

## Left as-is

**Atomic opt-out.** Making it atomic would mean relying on `get_uow` and `get_social_uow` happening to resolve to the same session through the dependency graph. They do today, but neither the interface nor the in-memory implementations guarantee it, so the atomicity would be real in production and absent in tests. The write order is already chosen so the failure mode is the harmless one: having stopped publishing with the history still up beats still publishing with the history already gone. A shared unit of work spanning both modules is the real fix and it is a bigger change than this PR should carry.

**History lookup loading history.** Fair, but it is bounded at 200 rounds and runs once when a round is closed, not per request. A dedicated `exists_completed_on_course` query on both repositories is the right shape and belongs with the feed work in #176, where the read paths get built anyway.

Full suite after the changes: 2540 unit, 427 integration and security, mypy clean, migration applied and rolled back against a real PostgreSQL.

PR #180, opened 2026-08-10
feat(social): Player profiles visible to friends, and the friends feed

agustinEDev:
Contributes to #176

Profiles, the friends feed, and the visibility rules around them.

## The visibility model changed mid-implementation

The issue as written said 404 for non-friends, reasoning that a 403 would confirm the account exists. That protection stopped making sense once it was decided that players are **searchable by name**: the account is discoverable by design, and hiding it would only prevent sending someone a friend request. [Decision recorded on the issue.](https://github.com/agustinEDev/RyderCupAM/issues/176#issuecomment-5246438957)

| | Any registered user | Friends |
|---|---|---|
| Find by name | yes | yes |
| Name, surname, photo | yes | yes |
| Handicap, stats, activity | no | yes |
| Send a friend request | yes | — |
| Remove the friendship | — | yes |

## Endpoints

```
GET /users/{id}/profile      # minimum card for anyone; full for friends
GET /users/{id}/activity     # friends only — 403 otherwise
GET /users/me/feed           # mine and my friends', cursor-paginated
PUT /users/me/feed/seen      # clears the unseen badge
```

## Decisions worth reviewing

**Private fields arrive as `null`, never zeroed.** A zero handicap reads as "plays terribly" rather than "you cannot see this".

**404 now means what it says** — the account does not exist or was deactivated. Activity answers **403**: a 403 leaks nothing once the client has just seen that player's card, and a fake 404 would only confuse it.

**The profile carries a friendship block** with `NONE` / `PENDING_SENT` / `PENDING_RECEIVED` / `ACCEPTED` / `BLOCKED` and the `friendship_id` needed to accept or remove it. The two pending states are separate because they do not lead to the same button: one waits on the other person, the other waits on you.

**Privacy is applied before paginating, not after.** Friends who turned sharing off are dropped from the author list *before* the query runs, so the page limit already applies to what can be shown. Filtering afterwards would return pages of irregular size — one of 20, the next of 14 — and leave the client unable to tell whether more remain. A test pins it: two friends with 5 achievements each, one of them silent, asking for 4 → 4 arrive, not 2.

**The feed includes your own achievements** but they never count towards `unseen_count`. A feed that omits what you just did reads as if it had not been saved, and it left the feed empty for anyone without friends yet — but being told "3 new things" about your own birdies is noise, not a notification.

**`find_friend_ids` is new** because the feed needs every friend at once. `list_friends` paginates, and half a friend list would have silently dropped the other half from the feed.

**The profile reuses `GetPlayerStatsUseCase`** rather than recomputing, so a friend's profile shows the same figures as that player's own dashboard instead of a second version that would drift.

## Security fix included

`GET /users/search-autocomplete` returned every matched user's **email address** to any authenticated caller, and the frontend rendered it (`AddFriendModal`, `SendInvitationModal`). Typing names harvested addresses of people you have no relationship with.

The email is gone. The endpoint now returns name, surname and avatar — the photo does the job the email was doing, telling apart two players with the same name — and an integration test fails if an address ever reappears in the response.

## A bug a test caught

`mark_feed_as_seen` was storing UTC. That value is compared against events' `occurred_at`, which the domain writes with `datetime.now()`, so the mark landed hours in the past and achievements published in that window would have counted as unseen **forever** — the badge would never fully clear. Fixed, with a test pinning the rule.

## Frontend note for #352

`SearchUsersItemDTO` no longer carries `email`; two screens render it today and will show blank until they switch to the avatar. `first_name` and `last_name` now arrive separated, so clients can stop splitting `full_name` on the first space — which breaks on compound surnames.

## Verification

- **2579** unit tests, **443** integration and security, 1 skipped
- `mypy` clean across 579 files; `ruff` clean on everything touched
- No migration in this PR — it reads what #175 already stores

## Summary by CodeRabbit

* **New Features**
  * Added player profile endpoints with friendship status, public details, and privacy-aware statistics.
  * Added player activity and friends feed views with cursor-based pagination.
  * Added feed “mark as seen” functionality and unseen activity counts.
  * Added privacy controls for activity and restricted profile information.
  * Improved user search results with names and avatar details while removing email addresses.
* **Tests**
  * Added comprehensive integration and unit coverage for profiles, feeds, activity visibility, privacy, pagination, and friendship behavior.

### DIFF
--- a/main.py
+++ b/main.py
@@ -57,6 +57,7 @@ from src.modules.quick_match.infrastructure.persistence.mappers.quick_match_mapp
     start_quick_match_mappers,
 )
 from src.modules.social.infrastructure.api.v1 import friend_routes  # noqa: E402
+from src.modules.social.infrastructure.api.v1 import profile_routes  # noqa: E402
 from src.modules.social.infrastructure.persistence.mappers.activity_event_mapper import (  # noqa: E402
     start_activity_event_mappers,
 )
@@ -432,6 +433,12 @@ app.include_router(
     friend_routes.router,
     prefix="/api/v1",
     tags=["Friends"],
+)
+
+app.include_router(
+    profile_routes.router,
+    prefix="/api/v1",
+    tags=["Profiles & Feed"],
 )
 
 app.include_router(
--- a/src/config/dependencies.py
+++ b/src/config/dependencies.py
@@ -266,9 +266,21 @@ from src.modules.social.application.ports.social_email_service_interface import 
     ISocialEmailService,
 )
 from src.modules.social.application.use_cases.block_user_use_case import BlockUserUseCase
+from src.modules.social.application.use_cases.get_friends_feed_use_case import (
+    GetFriendsFeedUseCase,
+)
+from src.modules.social.application.use_cases.get_player_activity_use_case import (
+    GetPlayerActivityUseCase,
+)
+from src.modules.social.application.use_cases.get_player_profile_use_case import (
+    GetPlayerProfileUseCase,
+)
 from src.modules.social.application.use_cases.list_friends_use_case import ListFriendsUseCase
 from src.modules.social.application.use_cases.list_pending_requests_use_case import (
     ListPendingRequestsUseCase,
+)
+from src.modules.social.application.use_cases.mark_feed_as_seen_use_case import (
+    MarkFeedAsSeenUseCase,
 )
 from src.modules.social.application.use_cases.publish_round_achievements_use_case import (
     PublishRoundAchievementsUseCase,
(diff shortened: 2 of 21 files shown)


agustinEDev:
Both findings were real and are fixed in `e9aefa0`.

**Reading your own activity with sharing off** returned an empty list, contradicting the feed — which already includes your own achievements regardless of that switch, because it governs what others see rather than what you see of yourself. The practical effect was worse than an inconsistency: turning sharing off left you unable to consult your own history. Now the switch only applies when the viewer is not the target, with a regression test.

**Cursors carrying a timezone offset** crashed pagination, and this one is client-reachable: the cursor is a query parameter, so anyone can send `+02:00` by hand and take the request down with a TypeError. Parsed cursors are converted to UTC and stripped of the offset, keeping the same instant on the scale everything else uses. Test added that pages with an offset-bearing cursor.

Suite after the changes: 2581 unit, 443 integration and security, mypy clean.

PR #185, opened 2026-08-11
chore(release): Prepare v2.7.0

agustinEDev:
The social component, plus half rounds counting towards the statistics.

`[Unreleased]` was empty despite seven commits since v2.6.0 — the fourth release in a row where merged PRs did not leave their entry — so the notes are written from `git log v2.6.0..develop`, not from what the block already said.

## What ships

- **Activity events** (#175) — six achievement types, published on closing a quick match or a tournament, with the opt-out switch that also removes what was already published.
- **Player profiles and the friends feed** (#176) — two visibility levels, cursor pagination, privacy applied before paginating. Includes the security fix that stopped user search returning email addresses.
- **Profile photo gallery** (#177) — twenty photos, friends only, proportions kept.
- **A friend's email on their profile** (#182).
- **Course names in the feed** (#183), resolved on read rather than stamped into the payload.
- **`share_activity` readable from the client**, so a settings screen can paint the switch in the state it is actually in.
- **Half rounds count** (#174), scaled to eighteen before mixing with the rest.

## Deploy order

This must reach production **before** the frontend release. The social pages there are useless without these endpoints.

Two migrations run on boot: `d4e8f1a3b7c9` (activity events) and `f7c2a9d4e6b8` (profile photos).

## Note for whoever reads `/health` after the deploy

`APP_VERSION` goes to `2.7.0` in this commit, so `curl https://api.rydercupfriends.com/health` is what confirms the deploy actually landed.


agustinEDev:
The gallery ships as **API only**: the endpoints are live and usable, but the frontend does not call them yet, so the feature does not exist from the outside until the next frontend release. The changelog entry now says so — as first written it read like a shipped user-facing feature.

Worth being aware of either way: live endpoints are reachable with a valid session regardless of whether any UI points at them. The exposure is bounded — friends-only visibility, a twenty-photo cap, authentication required — but this is the first release where photo upload is reachable in production.

PR #191, opened 2026-08-12
feat(golf-course): Add location data to golf courses

### LINKED ISSUE #105: [Feature]: Add location data (coordinates, address) to golf courses
## Description
Extend the golf course model with location data (address, latitude/longitude) to enable a future "courses near me" feature.

## Acceptance criteria
- [ ] Decide data source/format for coordinates (manual entry, geocoding on address input, etc.)
- [ ] Extend course model/migration with address + coordinates fields
- [ ] Expose location fields in the course API
- [ ] Backfill existing courses where possible

## Priority
P2 — Normal.

## Additional context
This is groundwork only — the "nearby courses" feature itself is out of scope for this issue. Also useful input for the Portugal/Spain course crawler.

agustinEDev:
## What

Golf courses can now store where they are: coordinates, postal address, city and province — all optional — across the domain aggregate, the API and migration `c4d8e1a72b93`.

This is the groundwork for suggesting nearby courses from the device location in quick matches. The proximity search itself is deliberately out of scope, as the issue states.

Closes #105

## Decisions worth reviewing

**Latitude and longitude go together or not at all.** Half a coordinate places nothing on a map, and a proximity search that accepted it would return arbitrary results. Enforced in three places — the value object, the DTO and a database CHECK — because the upcoming federated course import writes thousands of rows at once and a half-written coordinate would go unnoticed.

**Plain columns, not a composite value object.** `CourseLocation` is recomposed when read. Location is optional and `composite()` does not handle NULL, which is the rule already documented in `CLAUDE.md`.

**Updating without a location keeps the existing one.** Clearing it requires sending the object with every value null. The current edit form predates this feature and omits the field entirely, so treating the omission as a delete would wipe a course's coordinates on a rename. Same reasoning applies to update proposals: a clone inherits the original's location so that approving it never drops the coordinates.

**A course without location returns `null`, not an object with five nulls**, so a client can tell at a glance whether the course can be placed on a map.

## Testing

- 36 new tests: value object, aggregate, application flow, persistence and API
- Full unit suite: 2673 passing
- Migration applied, rolled back and reapplied against a scratch database; partial index and all three CHECKs verified in PostgreSQL
- `ruff check` and `mypy` clean

## Summary by CodeRabbit

* **New Features**
  * Added optional location details for golf courses, including coordinates, address, city, and province.
  * Locations are available when creating, viewing, and updating courses.
  * Updates preserve existing locations when omitted and remove them when explicitly cleared.
  * Added validation for coordinate ranges and paired latitude/longitude values.
* **Documentation**
  * Documented the new golf-course location support in the changelog.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/).
 
 ## [Unreleased]
 
+### Added
+
+- **Los campos de golf guardan dónde están** (issue #105): coordenadas, dirección postal, localidad y provincia, todo opcional, en el agregado, la API y la migración `c4d8e1a72b93`. Es la base para proponer campos cercanos a la ubicación del dispositivo al crear una partida rápida; la búsqueda por cercanía en sí queda fuera de esta entrega. **Latitud y longitud van juntas o no van**: media coordenada no sitúa nada en un mapa y una búsqueda por proximidad que la aceptara devolvería resultados arbitrarios. La regla la hacen cumplir el Value Object, el DTO y un CHECK en la base de datos, porque la importación masiva de campos federados escribe miles de filas de golpe y un dato a medias pasaría inadvertido. La ubicación se persiste en columnas sueltas y no como Value Object compuesto: es opcional, y `composite()` de SQLAlchemy no admite NULL. **Editar un campo sin enviar ubicación la conserva**; para borrarla hay que mandar el objeto con todos sus valores a `null`. Sin eso, el formulario de edición actual —que es anterior a esta funcionalidad y no manda el dato— habría borrado las coordenadas de un campo al cambiarle el nombre. Un campo sin ubicación la devuelve como `null` y no como un objeto con cinco nulos dentro, para que el cliente distinga de un vistazo si se sabe dónde está.
+
 ## [2.7.0] - 2026-08-11
 
 ### Added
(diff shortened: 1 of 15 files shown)


agustinEDev:
Thanks — two of the three findings were the same real bug and are fixed in 864572b.

**Fixed: zero coordinates treated as an empty location.** `CourseLocation.is_empty` tested truthiness, so a course at `(0.0, 0.0)` was reported as having no location and the API returned `location: null` for it. Now every field is compared against `None`. Covered by a test in the value object and another through the application flow, which is where the symptom showed.

**Not applying: requiring all five fields for a clear request.** The contract here is whole-object replacement, which is exactly what `tees` and `holes` already do in this same PUT — sending them replaces the collection rather than patching it. Making `location` the one field that needs an explicit five-null object would make the endpoint inconsistent with itself and add a rule a client cannot guess.

More to the point, the proposed validation does not address the case that actually risks data loss. `{"location": {"city": null}}` is a strange payload nobody sends by accident; `{"location": {"city": "Marbella"}}` is the realistic one, and it wipes the coordinates under either rule. The real fix for that would be per-field patching, which we considered and rejected for the inconsistency it introduces.

What was genuinely missing was that the contract was underdocumented, so the field description now states plainly that the object replaces the stored location in full, and a test pins the behaviour: a partial location clears the fields left out.

PR #194, opened 2026-08-12
fix(users): Exclude deactivated accounts from player search

### LINKED ISSUE #193: fix(users): Deactivated accounts still show up in player search
## Description

`search_by_partial_name` does not filter by `is_active`, so a deactivated account is still returned by the player search that powers "add a friend". The admin listing (`count_all` / `find_all`) does filter by it, so the two paths disagree about who is visible.

Anyone can search by partial name, so this exposes accounts that were explicitly deactivated and lets other users send them friend requests they will never see.

This also blocks the RFEG course import: the courses will be created by a system user, which would otherwise appear in the player search alongside real people.

## Acceptance criteria

- [ ] `search_by_partial_name` excludes inactive users in the SQLAlchemy repository
- [ ] Same behaviour in the in-memory repository, so tests do not diverge from production
- [ ] Tests covering an inactive user being excluded and an active one still found

## Priority

P2 — Normal.

agustinEDev:
## What

`search_by_partial_name` did not filter by `is_active`, while the admin listing does. The two paths disagreed about who is visible.

Any registered user can search by partial name, so a deactivated account was still exposed to everyone and could be sent friend requests it would never see.

Closes #193

## Why now

This also blocks the RFEG course import: those 802 courses will be created by a system user, which would otherwise show up in the player search alongside real people.

## Notes for review

**The filter runs before results count towards the limit.** Had hidden accounts consumed the quota, a legitimate account could fall out of a 10-result search because of accounts nobody should see. There is a test for exactly that.

**Fixed in the in-memory repository as well.** A test double that keeps returning deactivated accounts would vouch for behaviour production no longer has.

The repository interface docstring now states the guarantee, so the next implementation does not have to rediscover it.

## Testing

- 3 unit tests on the in-memory repository (exclusion, reactivation, limit accounting)
- 2 integration tests on the SQLAlchemy repository against PostgreSQL
- 2 end-to-end tests driving the real endpoints: an admin deactivates an account, another user searches for it
- Full unit suite: 2678 passing. `ruff` and `mypy` clean


agustinEDev:
@coderabbitai review

This PR was merged before you had reviewed it — your check reported `Review rate limited`. Please review it anyway: the change is already on `develop`, so anything you find will be fixed in a follow-up PR.

Issue #192, opened 2026-08-12
[Feature]: Import golf courses from Portugal


agustinEDev:
## Description

Import Portuguese golf courses into the platform, reusing the pipeline built for Spain (extract to JSON → map to DTO → importer with mandatory dry run).

Split out from #107, which is closed by the Spanish (RFEG) import. The domain model is already country-agnostic: `CourseSource` is a generic enum and `external_id` is nullable precisely so a second country can be added without another schema change.

## Sources

Two sources are needed — neither has everything, so they must be cross-referenced:

1. **Playing data** (course rating, slope, scorecards) — Federação Portuguesa de Golfe, WHS-classified courses:
   https://portal.fpg.pt/handicaps-course-rating/campos-classificados-whs/
   This is the Portuguese equivalent of what the RFEG publishes.

2. **Location and club data** (address, coordinates, contact):
   https://www.portugalgolf.net/en/golf-courses/
   Note: this is **not** the federation, it is a commercial portal. Check its terms of use and how current the data is before depending on it.

## Acceptance criteria

- [ ] Confirm the FPG source is scrapable and check terms of use for both sources
- [ ] Extractor producing the same JSON shape as the Spanish dataset
- [ ] Cross-reference the two sources by club — normalised name **and** geographic proximity, not text alone: in Spain 11 clubs name their course differently from the club itself, and Portugal will have the same problem
- [ ] Add `FPG` (or whichever origin applies) to `CourseSource`
- [ ] Import via the existing importer, with dry run and idempotency
- [ ] Handle courses that appear in one source but not the other

## Priority

P2 — Normal. Spain ships first.

Issue #107, opened 2026-07-25
[Feature]: Build crawler to import golf courses from Spain


agustinEDev:
## Description
Build a crawler/importer to source golf course data (name, address, holes, par, etc.) for Spain and load it into the platform.

## Acceptance criteria
- [x] Identify source(s) for course data in Spain
- [x] Crawler/import script producing data matching the course model (including location fields, see the "Add location data to golf courses" issue)
- [x] De-duplication against existing courses
- [x] Idempotent, re-runnable import

## Priority
P2 — Normal.

agustinEDev:
Scope reduced to Spain. Portugal was split out into #192, which carries the sources and the acceptance criteria for it.

The Spanish import is delivered by #196: 802 RFEG courses with 4,304 tees and 77,472 holes, created through the use cases, with a dry run by default, deduplication against existing courses and a re-runnable import.

PR #195, opened 2026-08-12
feat(golf-course): Add provenance and physical holes to golf courses

agustinEDev:
## What

Golf courses now record where their data comes from — source, external id, import date — plus how many holes the course actually has on the ground. Migration `e7f2b3c9d418`. Also adds the system account that will own the imported courses.

Groundwork for the RFEG importer. Part of #107, which closes with the importer itself.

## Decisions worth reviewing

**Source is an enum, not free text.** `MANUAL` and `RFEG` today; each new federation is one more value. Free text would let 'RFEG', 'rfeg' and 'R.F.E.G.' coexist meaning the same thing.

**The external id is deliberately nullable.** Not every federation publishes a stable id per course, and the model has to accept countries whose source has none — Portugal is already queued as #192.

**The unique index is partial.** Two courses cannot be the same course from the same source, but manual courses carry no external id and would all collide with each other in a plain unique index. It is declared on the table and not only in the migration, because it is a correctness rule and the schema the tests build from metadata needs it too — the duplicate-import test failed until that was fixed.

**Provenance is not accepted in API requests.** It goes as a use case parameter, not in the request DTO: a client cannot claim its data is endorsed by a federation. Editing a course does not alter it either, and a clone inherits it, so approving a rename does not silently turn a federated course into a manual one.

**`physical_holes` is 9 or 18, and null means unknown.** Federations always publish an 18-hole scorecard, so a nine-hole course appears with the round repeated. Null is honest for every course created before now: nobody was ever asked.

**The system account is created deactivated.** It cannot log in, and since #193 it stays out of player search. Not an admin — it executes nothing, it only shows up as the author. Its password is random and nobody knows it. Obtaining it is idempotent, so re-importing does not create a second one.

## Testing

- 28 new tests: two value objects, the aggregate, the system account use case and persistence
- Full unit suite: 2706 passing
- Migration applied, rolled back and reapplied against a scratch database; enum, both CHECKs and the partial unique index verified in PostgreSQL
- `ruff` and `mypy` clean

Note: `ruff check` also reports 5 pre-existing `RUF100` warnings in `test_quick_match_endpoints.py`, untouched here — the local ruff is newer than CI's.

## Summary by CodeRabbit

* **New Features**
  * Added golf-course provenance details, including data source, external identifier, and import date.
  * Added physical hole-count support for 9- or 18-hole courses, with unknown values supported.
  * Preserved provenance and hole-count information through course updates and proposals.
  * Added inactive, non-administrator system accounts for imported course data.
* **Bug Fixes**
  * Prevented duplicate imported courses with the same source and external identifier.
  * Added validation for required import dates and supported hole counts.


agustinEDev:
Two of the four findings were real and are fixed in c8f8d96. The other two I am not applying, with reasons.

**Fixed: the source enum was not enforced.** `CourseProvenance` is a dataclass, so the `CourseSource` annotation is documentation, not validation. A lowercase `"rfeg"` passed every rule precisely because it is not `CourseSource.MANUAL`, and would have failed much later against the PostgreSQL enum column or when serialising the response. `__post_init__` now converts through `CourseSource(...)`, so an unknown or non-canonical source is rejected where the error still makes sense. Covered by tests for `"rfeg"`, `"FPG"`, `""`, `"MANUAL_2"` and for a lowercase `"manual"` smuggling in import data.

**Fixed: the system account identity was not verified.** Returning any account holding that email meant that, had someone registered with the address, hundreds of imported courses would have been attributed to a personal account. It now fails closed if the account is active or an administrator, with a message naming the address. Two tests cover it.

**Not applying: a sentinel to distinguish an omitted `physical_holes` from an explicit unknown.** The distinction is real, but it buys a transition the product does not have. Going from `9` back to "unknown" is not a correction anyone would make: once you know whether a course is nine or eighteen holes, you fix a wrong mark by setting the right value, not by returning to ignorance. Unlike `location`, where the ambiguity risked silently wiping coordinates, here the default is the safe one — omitting preserves. A private sentinel would leak into the signature of `update()`, `apply_update()` and clone creation to serve a case that never arrives. The contract is now stated explicitly in the docstring instead.

**Not applying: handling the concurrent-creation race on the system account.** The importer is a script run by hand, once, from a single process; there is no second writer to race against. Catching the unique-constraint conflict and reloading would add a branch that cannot be exercised by any real caller, and an untested branch in the path that decides who owns imported data is worse than not having it. If the account is ever created from a concurrent context, the unique constraint on `users.email` still prevents duplicates — the import would fail loudly rather than attribute data wrongly, which is the outcome that matters.

PR #198, opened 2026-08-12
Search, paginate and sort golf courses by distance

### LINKED ISSUE #197: [Feature]: Light golf course listing with search, pagination and proximity
## Description

The golf course list endpoint returns **every** approved course in one response. It already omits the per-tee scorecards (`include_tee_scorecards=False`), but it still carries the course-level `holes` and every tee. With the 802 RFEG courses imported that is 14,436 hole objects and 4,304 tees in a single payload of roughly 1.6 MB, serialised just to paint a list of names. The frontend then filters by name **in the browser** (`GolfCourseSearchBox.jsx`), so the whole catalogue travels on every visit to the selector.

Three changes, all on `GET /api/v1/golf-courses`:

**1. Search and pagination — this is where the size goes.** Partial name search and `limit`/`offset`, filtered in the database rather than in the browser. Sending 20 courses instead of 802 is what actually fixes the payload; the DTO shape is a secondary saving.

**2. Light listing DTO.** Drop the course-level `holes` — roughly half of what remains, and nothing in the listing path reads them. Keep the tees with `color`, `identifier`, `tee_gender`, `course_rating` and `slope_rating`. The full course stays available at `GET /golf-courses/{id}` for when one is actually selected.

**3. Proximity, for quick matches.** Optional `lat`/`lon` that sort by distance and return it per result, with an optional `radius_km` cap. In a quick match the player is standing on the course, so distance is the natural ordering.

## Why the tees stay

They are about 30% of the payload, so dropping them is tempting. Two consumers would break:

- `GolfCourseTable` renders one colour badge per tee (`course.tees.map`), used by the admin panel.
- The `Tee` value object **validates in its constructor**: `courseRating` and `slopeRating` must be finite numbers, and colour `OTHER` requires an identifier. Sending trimmed tees makes the frontend throw when building the entity, it does not degrade. `holes` defaults to `[]`, so leaving those out is safe.

With pagination in place the tees cost about 2 KB per page, so they are not worth the breakage.

## Frontend change this forces

`pages/admin/GolfCourses.jsx` feeds the edit form straight from the listing (`handleOpenEdit` → `initialData={selectedCourse}`). `GolfCourseForm` only loads holes `if (initialData.holes.length === 18)`, and its default state is 18 holes at par 4 with sequential stroke indices. With a light listing the admin could open a course, save, and **silently overwrite a real scorecard with 18 pars of 4**.

The form must be fed from `GET /golf-courses/{id}`, the way `CreateQuickMatchModal` already does after selection. **The light backend must not be deployed without this frontend fix.**

## Proximity: why not PostGIS

Not available in the cluster's Postgres image; `cube` and `earthdistance` are present but not installed. At this scale a haversine in SQL with a bounding-box prefilter is instant, needs no extension and behaves identically in local and on Render.

A fixed radius behaves very differently by region — measured against the imported data, within 50 km: Madrid 69 courses, Málaga 53, Barcelona 37, Bilbao 13, Soria 3. Hence sorting by distance and returning the nearest N, with the radius as an optional cap rather than the main criterion: it degrades gracefully where there is nothing nearby.

12 of the 803 courses have no coordinates and are invisible to proximity search, so name search must always remain available — also for when the user denies geolocation.

## Acceptance criteria

- [ ] Partial name search, filtered in the database
- [ ] Pagination with a total count
- [ ] Light listing DTO without the course-level holes, tees intact
- [ ] Optional `lat`/`lon` ordering by distance, distance returned per result
- [ ] Optional `radius_km` cap
- [ ] Courses without coordinates excluded from proximity results, never from name search
- [ ] Admin edit form fed from the detail endpoint (frontend)
- [ ] Selector queries the backend instead of filtering the full catalogue in the browser (frontend)

## Priority

P1 — the selector is unusable on mobile with 802 courses.

agustinEDev:
## What

`GET /golf-courses` returned **every** approved course in one response, and the frontend filtered by name in the browser. With the 802 imported courses that is roughly 1.6 MB on every visit to the selector, and the filtering runs on the user's phone instead of in the database.

The endpoint now takes:

| Parameter | Effect |
|---|---|
| `name` | Partial match, case-insensitive, filtered in SQL |
| `limit` / `offset` | Pagination. The response reports `total` so the client knows whether to ask for more |
| `lat` / `lon` | Sorts nearest first and returns `distance_km` per course |
| `radius_km` | Optional cap. Requires a position |

`limit` is optional on purpose: without it everything comes back, exactly as before. A client that does not paginate yet must not stop seeing courses because the server changed.

The listing no longer carries the course scorecard — 802 courses by 18 holes were 14,436 objects travelling to paint a list of names. It is now its own `GolfCourseSummaryDTO` rather than the full response DTO, so the listing contract can move without dragging the detail one along.

Closes #197 for the backend half. The frontend is a separate PR and **must ship with this one** — see below.

## Why sort by distance instead of cutting at 50 km

Measured against the imported data, courses within 50 km:

| | 25 km | 50 km | 100 km |
|---|---|---|---|
| Madrid | 54 | 69 | 85 |
| Málaga | 25 | 53 | 120 |
| Barcelona | 18 | 37 | 60 |
| Bilbao | 10 | 13 | 47 |
| Soria | 1 | 3 | 14 |

A fixed radius gives an unusable 69-item list on the meseta and an empty one across half the country. Sorting by distance and returning it lets the client say "the nearest is 78 km away" instead of showing nothing.

## Why not PostGIS

Not in the cluster's Postgres image; `cube` and `earthdistance` are available but not installed. At this scale the haversine in SQL costs less than keeping an extension installed and migrated across three environments. If the catalogue ever grows by an order of magnitude this is the thing to revisit.

## The tees stay, and that is deliberate

They are about 30% of what remains, so dropping them is tempting. Two things break if we do:

- `GolfCourseTable` renders a colour badge per tee, used by the admin panel.
- The frontend `Tee` value object **validates in its constructor** — ratings must be finite numbers. Trimmed tees make it throw when building the entity; it does not degrade.

With pagination in place they cost about 2 KB per page, so the breakage is not worth it.

## ⚠️ This must not be deployed without the frontend PR

`pages/admin/GolfCourses.jsx` feeds the edit form straight from the listing (`handleOpenEdit` → `initialData={selectedCourse}`). `GolfCourseForm` only loads holes `if (initialData.holes.length === 18)`, and its default state is 18 holes at par 4 with sequential stroke indices.

Without the scorecard in the listing, an admin can open a course, change the name, save, and **silently overwrite a real scorecard with 18 pars of 4**. On 802 freshly imported federation courses.

The fix is three lines — fetch the course by id before opening the modal, the way `CreateQuickMatchModal` already does after selection — but it lives in the frontend repo.

## Two traps worth reviewing

**`joinedload` → `selectinload` for the tees.** The join multiplies rows, so `LIMIT 20` would have cut a course's tees in half rather than returning 20 courses. There is a test that asserts every course in a page still has all its tees.

**`acos` is clamped to [-1, 1].** Floating point reaches 1.0000000000000002 for the course you are standing on, and that is a domain error in PostgreSQL — it would take down the whole query. Tested by searching from a course's exact coordinates.

## Testing

- 11 new unit tests on the use case, driven through the **in-memory repository rather than a mock**, so the double actually applies the filters
- 7 new integration tests on the repository: name search, wildcard escaping, pagination keeping tees intact, distance ordering, radius, courses without coordinates, and the exact-coordinates case
- 4 new API tests: `count`/`total`, half a coordinate → 400, radius without position → 400, oversized limit → 422
- Full unit suite: 2808 passing. `ruff` and `mypy` clean

Integration tests were not run locally — the cluster is down and the API image needs rebuilding. CI covers them.

https://claude.ai/code/session_01HH1iy1idhKo15kc2majyyu

## Summary by CodeRabbit

* **New Features**
  * Added golf-course search by partial name.
  * Added optional pagination with result counts and total matching courses.
  * Added proximity sorting and radius filtering using latitude and longitude.
  * Listing results now provide course summaries with tees; full course details remain available through the course details endpoint.
  * Pagination remains optional for backward compatibility.

* **Bug Fixes**
  * Added clear validation errors for incomplete coordinates, radius-only searches, and invalid limits.


agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai review

Issue #200, opened 2026-08-12
[Feature]: Edit a course scorecard per tee, metres included


agustinEDev:
## Description

The course edit form describes a course as "18 holes, plus some tees with ratings". That stopped being what a course is.

Since the RFEG import, a course is **N tees, each with its own scorecard** — par, stroke index and metres per hole, per tee. The course-level card is a *derived* view of the first tee, kept so that consumers reading `golf_course.holes` keep working. Editing hole by hole at course level no longer describes the object: it cannot express that the yellow tee plays 350 m where the white plays 385 m, which is the whole point of having tees.

The form should let you pick a tee and edit **its** card, metres included.

Note this is the redesign, not the bleeding. The immediate data loss is #199 and must land first — it is a backend fix and does not depend on this.

## The backend half: the update contract is inverted

Worth knowing before designing the screen, because it decides what the screen can send.

There are only three tables: `golf_courses`, `golf_course_tees` and `golf_course_tee_holes`. **There is no course-level holes table.** `GolfCourse.holes` is derived — rebuilt from the first tee on load, as the mapper says at line 315. The scorecard belongs to the tee, and that is the only thing persisted.

`UpdateGolfCourseRequestDTO` has it the other way round: it **requires** 18 course-level holes and treats the per-tee cards as optional. The derived view is mandatory and the source of truth is optional. That inversion is what allowed #199 to happen — a course-level card, arriving without metres, overwriting the real ones.

#199 stops the bleeding by treating an omitted tee card as "leave it alone", which is the right rule on its own terms and stays true regardless of what happens here. But it does not fix the contract.

What the contract should become: an edit sends **the tees with their scorecards**, and the course-level card is derived from them and never accepted from a client. That is the backend half of this issue, and it is what lets the screen exist at all — today it has no way to send metres even if it collected them.

Note the courses where the tees encode the layout: the import found **83 nine-hole courses that play the second round from different tees**. For those, the per-tee cards are not a refinement, they are the structure of the round. Any contract that keeps treating the course card as primary loses them.

## What the form does not cover today

- Metres per hole. Never collected, in any form, so any edit sends the card without them.
- Stroke index per tee. Only one is collected, at course level.
- Which tee a card belongs to. There is a single card and it is applied to all of them.

## Open questions worth deciding before building

[...]

PR #201, opened 2026-08-12
Keep each tee scorecard when an edit omits it

### LINKED ISSUE #199: [Bug]: Editing a course wipes the per-tee scorecards
## What happens

Editing a golf course **wipes every tee's own scorecard** and replaces it with the course-level reference card, which itself arrives without metres. A course imported from the RFEG with five tees of different lengths ends up with five identical tees and no distances at all.

This is live on `develop` today. Any admin edit of any of the 802 imported courses destroys their geometry. The ratings survive; the scorecards do not.

## The chain

1. `GolfCourseForm` only collects colour, identifier, gender and the two ratings per tee. It never collects metres, nor a stroke index per tee.
2. `ApiGolfCourseRepository._mapToApiPayload` sends the tees **without `holes`**, and the course card with only `hole_number`, `par` and `stroke_index` — no `meters`.
3. `GolfCourseMapper.to_domain_tees` builds each tee with `holes=[]` when the DTO carries none.
4. `GolfCourse._sync_holes_and_tees` then does exactly what it documents: it copies the course card onto every tee that has no card of its own.

Step 4 is correct for a **new** course described by a single card. It is wrong for an **update** that simply does not mention the tee scorecards, and the aggregate cannot tell the two apart.

## Why it matters now

Before the RFEG import, a course was "18 holes plus some tees with ratings", so copying the course card onto the tees lost nothing. Now the federation publishes a card per tee — 4,304 tees with their own metres and stroke indices — and the course-level card is a *derived* view of the first tee, not the source of truth.

The data is not hard to get back for imported courses: re-running the importer restores them. For a course edited by hand there is no source to restore from.

## Suggested fix

Make an update that carries no per-tee scorecard **keep the existing one** rather than overwrite it. The distinction the aggregate needs is between "this tee has no card because the course is described by a single one" and "this update does not mention the cards", which is information the update use case has and the constructor does not.

Worth considering alongside: whether the admin panel should refuse to edit the scorecard of a course whose provenance is `RFEG` at all. Those values are the federation's, and the next import would put them back anyway, which is more confusing than blocking the edit.

## Acceptance criteria

- [ ] An update that omits the per-tee scorecards leaves them untouched
- [ ] An update that provides them replaces them
- [ ] Creating a course with a single card still copies it to the tees that have none
- [ ] Regression test with a course whose tees have different metres, edited without mentioning the cards

## Priority

P1 — silent data loss on already imported data. It needs to land before any frontend work on the edit form; the redesign of that form is tracked separately.

agustinEDev:
## What

Editing a golf course wiped every tee's own scorecard. On `develop` today, any admin edit of any of the 802 imported courses destroys their geometry — silently, no error.

The edit form predates per-tee scorecards. It sends the tees with their ratings but **with no holes**, and the aggregate read that omission as "these tees have no card of their own" and copied the course reference card onto them — a card that itself arrives without metres, because the form has never asked for them.

So editing the name of a federated course turned its five tees of different lengths into five identical ones with no distances, and levelled their stroke indices along the way. Stroke allocation depends on those.

An edit that does not mention a tee's scorecard now keeps the one it had, the same way it already works for the location: **omitting is not deleting**.

Closes #199.

## The chain, for review

| Step | Where |
|---|---|
| Form collects colour, gender, identifier and ratings. Never metres | `GolfCourseForm.jsx` |
| Payload sends tees without `holes`, course card without `meters` | `ApiGolfCourseRepository._mapToApiPayload` |
| Each tee is built with `holes=[]` | `GolfCourseMapper.to_domain_tees` |
| The course card is copied onto every tee without one | `GolfCourse._sync_holes_and_tees` |

The last step is correct for a **new** course described by a single card. It is wrong for an **update** that simply does not mention the cards, and the constructor cannot tell the two apart. The update path can.

## How a tee is recognised

By colour, gender and identifier. Colour alone is not enough: a course can publish two `OTHER` tees of the same gender — the British "Championship", the American combined tees — and the identifier is the only thing separating them. It is normalised (trimmed, case-folded) so that retyping a name with different capitalisation does not turn an existing tee into a new one and lose its metres.

Three cases, all tested:

- Tee **without** a card → keeps the one it had
- Tee **with** a card → imposes it, which is how you actually edit one
- **New** tee that did not exist → inherits the course card, as it would on creation

## Where the fix lives

In `apply_update`, before it dispatches between the in-place edit and the update proposal. Both paths needed it: a creator editing an approved course does not update in place, it creates a clone for an admin to approve. A clone born with the cards already lost would have wiped them on approval.

`update()` itself is left alone and now documents that it takes the tees literally. It is the primitive; `apply_update` is where edits actually enter.

## Testing

11 new unit tests on the aggregate. **Six of them fail without the fix** — verified by disabling the line and re-running, not assumed:

- metres survive an edit that only changes the name
- the tees do not end up identical to each other
- a tee that brings a card imposes it
- a new tee takes the course card
- gender is part of the identity
- the identifier tells two `OTHER` tees apart
- a change of capitalisation does not lose the card
- the update proposal keeps the cards too
- editing the ratings still works
- the stroke index survives as well

Full unit suite: 2808 passing. `ruff` and `mypy` clean. No migration, no frontend change.

## What this does not fix

The form still cannot **edit** metres — it never collects them. That is #200, the redesign that lets you edit a scorecard per tee. This PR only stops the destruction.

https://claude.ai/code/session_01HH1iy1idhKo15kc2majyyu

## Summary by CodeRabbit

* **Bug Fixes**
  * Editing a golf course now preserves existing tee-specific scorecards when no replacement card is provided.
  * Explicitly supplied scorecards replace only the selected tee’s card.
  * Newly added tees correctly inherit the course scorecard.
  * Tee matching is more accurate across color, gender, and identifiers, preventing cards from being assigned to the wrong tee.
  * The fix applies to both direct course edits and proposal-based updates, while preserving rating changes.


agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai review

Issue #207, opened 2026-08-15
A quick match's play mode cannot be changed after creation


agustinEDev:
## What happens

`play_mode` (SCRATCH / HANDICAP) is set when the quick match is created and there is no endpoint to change it afterwards. Pick the wrong one and the only way out is to delete the match and build it again — re-adding every participant, their tees and their handicaps.

## Why it matters

It is easy to get wrong and expensive to undo. The mode sits next to the allowance in step 1 of the wizard, before the players are even added, and it changes everything downstream: whether strokes are given at all, what the card paints, and who wins each hole.

It also constrained the migration that introduced the field. Existing match play rows had been resolved gross, which is exactly SCRATCH, but only the **finished** ones were backfilled that way — a PENDING or IN_PROGRESS match had to be left on HANDICAP, because marking it SCRATCH would have been permanent and its players had set it up expecting handicaps. With an endpoint, that call would have been the creator's rather than the migration's.

## Scope

- `PATCH /api/v1/quick-matches/{id}/play-mode`, creator only, PENDING or IN_PROGRESS only.
- Worth deciding: allow it mid-round, or only before the first score is recorded? Changing it after a few holes rewrites the standing of holes already played. Suggest allowing it while PENDING and blocking it once a score exists, which is easy to explain.
- Frontend: surface it wherever a quick match is edited, and say plainly what it changes.

## Related

The same limitation applies to `allowance_percentage`, which is also create-only. Worth solving together.

Came out of `fix/quick-match-handicap-scoring`, where the play mode was introduced.

Issue #206, opened 2026-08-15
Pitch & putt courses fall back to the raw Handicap Index


agustinEDev:
## What happens

`TeeRating` validates `66 <= par <= 76` (`competition/domain/services/playing_handicap_calculator.py`). A federated pitch & putt is par ~54, so **every one of its tees raises `ValueError`**, gets skipped by `StrokeContextBuilder`, and all participants fall back to playing off their raw Handicap Index instead of a Course Handicap.

The same bound rejects `EXECUTIVE` courses.

## Why it matters

It is silent. The match is playable and the numbers look plausible, they are just not the WHS ones. A player off 18 gets 18 strokes instead of the ~0 that a par-54 course off SR 47 / CR 46.8 would give them — the allocation is wrong by the whole handicap.

It also splits the two implementations in a way we deliberately closed elsewhere: the frontend `Tee` value object accepts CR 45-90 and SR 40-160 precisely so these courses can be stored, so `MatchPlayStrokeAllocator` had to mirror the *backend's* narrower bounds to stay in parity (`RATED_MIN_*` constants). Widening one side without the other reopens the drift.

## Why it was not fixed with the rest

`TeeRating` is shared with `competition`. Widening it changes the playing handicaps of competitions that already exist, and their matches are generated once and stored — so this needs its own change with a decision about existing data, not a line in a bugfix branch.

## Suggested approach

1. Decide the real WHS bounds for short courses (pitch & putt, executive) and widen `MIN_PAR` / `MAX_PAR`, `MIN_COURSE_RATING`, `MIN_SLOPE_RATING` accordingly.
2. Work out whether already-generated competition matches need regenerating, or whether the change applies only to new ones.
3. Relax the mirrored `RATED_*` constants in `MatchPlayStrokeAllocator` (RyderCupWeb) to match, and regenerate `scripts/generate_parity_fixtures.py` so the parity test covers a short course.

## Where to look

- `src/modules/competition/domain/services/playing_handicap_calculator.py` — `TeeRating.__post_init__`
- `src/modules/quick_match/application/services/stroke_context_builder.py` — the `except ValueError` that skips the tee (it logs a warning once per course, so production logs will show which courses are affected)
- `RyderCupWeb/src/domain/services/MatchPlayStrokeAllocator.js` — `isRatable`

Found while reviewing the quick match stroke allocation branch (`fix/quick-match-handicap-scoring`).

agustinEDev:
[...]

PR #208, opened 2026-08-15
Fix match play stroke allocation in quick matches and competitions, and add a scratch mode

### LINKED ISSUE #209: Competition stroke allocation uses the reference tee's card for every player
Audit of the `competition` scoring path, prompted by the quick match fixes in #208. Most of that family of bugs is **not** present here — the two that are, are below.

## What is already correct in `competition`

| Bug found in `quick_match` | Present in `competition`? |
|---|---|
| Result computed gross instead of net | **No.** All four callers of `calculate_hole_winner` pass `hs.net_score` |
| Match play allocating the full playing handicap | **No.** Singles, fourball and foursomes all use the differential |
| Inconsistent rounding | **No.** Everything goes through `PlayingHandicapCalculator` / `ROUND_HALF_UP` |
| Missing `(colour, None)` tee fallback | **No.** `_resolve_player_data` has it — it is where the quick match fix was copied from |
| Plus handicaps | **N/A.** Competition is match play only; the differential covers it and clamping at zero is correct there |

## Bug 1 — strokes are dealt on the reference tee's holes

`generate_matches_use_case.py:242`

```python
holes_by_stroke_index = [
    h.number for h in sorted(golf_course.holes, key=lambda h: h.stroke_index)
]
```

`golf_course.holes` is only the **first tee's** card (see `GolfCourse._sync_holes_and_tees`). The RFEG importer stores a card per tee, and each can carry its own stroke index. A player on a tee whose ordering differs from the reference gets their strokes on the wrong holes.

## Bug 2 — every tee is rated against the course par

`generate_matches_use_case.py:235`

```python
total_par = sum(h.par for h in golf_course.holes)
```

That course par is then used to build `TeeRating` for **all** tees, instead of each tee's own `par_total`. Where a tee's par differs, `CR - par` is off and so is the course handicap.

## Exposure, measured on the imported dataset

Of the 800 federated courses with more than one carded tee:

- **56 (7.0%)** have a stroke index that differs between tees
- **25 (3.1%)** have a par that differs between tees
- **80 (10.0%)** are hit by at least one

At Aloha, every ladies' tee differs from every men's — so any mixed competition there allocates wrongly.

## Why this one is worse than the quick match version

Competition matches are **generated once and persisted**: `strokes_received` is stored per `MatchPlayer` at generation time. A wrong allocation is frozen in the database and does not correct itself when the code is fixed.

So this needs a decision, not just a patch: leave existing competitions alone, or regenerate the matches of the affected ones (only for rounds that have not started, presumably).

## Bug 3 — `handicap_strokes_given` is doubled in foursomes (display only)

`match.py:110-116` sums the `playing_handicap` of each team's players, but in foursomes both players carry the **same team** handicap, so the total comes out doubled. Reproduced:

```
Team A CH [13, 16] · Team B CH [26, 31], 50% allowance
Real strokes to team B:            7
handicap_strokes_given on the card: 14
```

Shown by `MatchCard.jsx:103` and `MatchDetailModal.jsx:157`. **Verified that the field feeds no calculation** — the per-player `strokes_received` is correct, so results are unaffected. It is the card contradicting itself, nothing more. In fourball the same sum produces a number that does not mean anything either (a sum of individual differentials).

Deriving it from the receiving side's `len(strokes_received)` would be right for all three formats. Note the column is persisted, so existing rows would need a backfill or would keep the old number.

## Suggested split

1. Bugs 1 and 2 together — they are the same fix in the same method, and share the question of what to do with existing competitions.
2. Bug 3 on its own — display only, and it needs a backfill decision for `handicap_strokes_given`.

The equivalent fix for quick matches is in #208, in `StrokeContextBuilder` (`holes_by_stroke_index_by_tee`) and `StrokeAllocationService._holes_for`, and can be reused.

agustinEDev:
## The bug

A 1 vs 1 match play quick match was **resolved gross**. `_compute_standing` fed the raw `hole_score.score` into `ScoringService.calculate_hole_winner`, whose contract is net scores, so a match between an 18.0 and a 20.7 was decided as if both played scratch. `GetRecentMatchesUseCase` did the same, on its own copy.

The strokes drawn on the card were wrong on top of that: they came from each player's full Playing Handicap, which is the stroke-play method. WHS match play gives only the *difference*, to the higher handicap, on the lowest stroke index holes. The totals looked close but landed on different holes — which is exactly where tight holes are decided.

Reported from a real match at Golf de Meis, where the reporter (18.0) was drawn **more** strokes than his opponent (20.7).

## Quick matches

- **`StrokeAllocationService`** (new, quick_match domain) resolves strokes per participant for every format: singles / fourball / foursomes differential, free play individual, nothing at all in scratch. The card and the result finally share one calculation.
- **`StrokeContextBuilder`** (new, application layer) translates a `GolfCourse` into what that service needs, keeping the domain free of `golf_course` entities.
- Detail and history now compare **net** scores, and the detail exposes `participant_strokes` so the frontend can draw what actually decided the holes.
- **Scratch mode**: quick matches gain `play_mode`, reusing the `PlayMode` value object competitions already use. Free play honours it too. Defaults to HANDICAP, so a caller that omits the field behaves exactly as before.

### Migration

`b4d7e1a9c25f` backfills existing rows to preserve what the app was showing rather than rewriting finished matches:

- match play that is **COMPLETED or CANCELLED** → `SCRATCH` (it was resolved gross, so its stored result does not move)
- everything else → `HANDICAP`, including match play still PENDING or IN_PROGRESS: those have no result to preserve and their players set them up expecting handicaps

## Competitions (closes #209)

An audit of the competition path found most of that family **absent** — every caller of `calculate_hole_winner` already passes `net_score`, all three formats already use the differential, rounding already goes through ROUND_HALF_UP, and the genderless-tee fallback was already there. Two defects were present:

- Strokes were dealt against `golf_course.holes`, which is only the first tee's card. Each tee can carry its own: **56 of the 800** federated courses with more than one carded tee have a stroke index that changes between tees.
- Every tee was rated against the course par instead of its own, shifting `CR - par` and the course handicap. **25 of those 800** have a par that differs between tees.

10% of federated courses are exposed to at least one. `GenerateMatchesUseCase` and `ReassignMatchPlayersUseCase` each had their own copy of the translation and both carried both defects, so the fix is a shared `TeeContextBuilder`, the same shape as `StrokeContextBuilder` in quick_match.

Also: `Match.create` summed each team's `playing_handicap`, which comes out **doubled** in foursomes where both partners carry the same team handicap — a team receiving 7 strokes was shown as 14. Verified the field feeds no calculation (only `MatchCard` and `MatchDetailModal` read it), so no result was ever wrong from this.

### Data already in the database

Competition matches are generated once and **persisted**, so a wrong allocation is frozen and does not correct itself on deploy.

`scripts/regenerate_scheduled_round_strokes.py` recomputes rounds still in `SCHEDULED`, **preserving pairings and match numbers**: it reads the existing pairings, reopens the round via the new `Round.reopen_for_regeneration` and re-runs generation with those pairings as manual ones, so the allocation comes out of the corrected code with no logic duplicated. Run `--dry-run` first.

Rounds that have started are deliberately left alone: they have results, and changing their strokes would rewrite holes already played.

## Review passes

An architecture/OWASP audit and a code review were run over this branch. Findings worth knowing:

- Plus handicaps were clamped at zero, so the card and the classification disagreed on screen. `ParticipantStrokes` now carries a **signed** hole → strokes map, which a list of hole numbers could not represent.
- The quick match tee lookup had no `(colour, None)` fallback — copied from competition, which did have it.
- The no-tee fallback dropped the allowance and rounded half-to-even, both out of step with the rest of the handicap calculation.
- `play_mode` was ignored by the Stableford and to-par paths, so a scratch round was credited in the history and the player average as if strokes had been given.

## Parity with the frontend

`scripts/generate_parity_fixtures.py` runs the quick match service over 12 scenarios and writes the exact results where the frontend's parity test reads them. It caught three live divergences the first time it ran. The companion frontend PR consumes it.

## Verification

- 2859 unit tests pass
- `ruff` clean, `mypy` clean over 605 files
- Single Alembic head

## Known limitations, tracked separately

- #206 — pitch & putt courses fall back to the raw Handicap Index (`TeeRating` rejects par 54)
- #207 — `play_mode` cannot be changed after creation

## Merge order

**Merge this before the frontend PR** (agustinEDev/RyderCupWeb#399). The frontend consumes `participant_strokes`; the reverse order leaves the card recomputing locally, which still works but is the situation this branch exists to end.

https://claude.ai/code/session_015Y4Jo33BPFFPYz7oDb8vaM

## Summary by CodeRabbit

- **New Features**
  - Quick matches now support **Handicap** or **Scratch** play modes.
  - Match details show each participant’s playing handicap and strokes by hole.
  - Handicap strokes use tee-specific ratings and stroke indexes.
  - Standings and recent-match results account for net scores where applicable.
  - Scheduled-round strokes can be regenerated with dry-run reporting.

- **Bug Fixes**
  - Scratch matches no longer apply handicaps to statistics or standings.
  - Improved handling of missing or invalid tee and handicap data.

- **Tests**
  - Expanded coverage for play modes, stroke allocation, standings, and regeneration.

### DIFF
--- a/alembic/versions/b4d7e1a9c25f_add_play_mode_to_quick_matches.py
+++ b/alembic/versions/b4d7e1a9c25f_add_play_mode_to_quick_matches.py
@@ -1,0 +1,66 @@
+"""add play_mode to quick_matches
+
+Anade el modo de juego (SCRATCH / HANDICAP) a las partidas rapidas, el mismo
+Value Object que ya usan las competiciones.
+
+Las filas existentes NO se rellenan todas igual, y es deliberado: se les asigna
+el modo que reproduce lo que la aplicacion venia mostrando, para no reescribir
+el resultado de partidas ya jugadas.
+
+- Partidas de match play (match_format IS NOT NULL): el resultado se venia
+  calculando con los golpes brutos, sin aplicar handicap. Eso es exactamente
+  SCRATCH, asi que se quedan en SCRATCH y su resultado no se mueve.
+- Partido libre (scoring_format IS NOT NULL): la clasificacion ya se calculaba
+  neta, aplicando el Playing Handicap de cada uno. Eso es HANDICAP.
+
+Las partidas nuevas nacen en HANDICAP salvo que el creador elija scratch.
+
+Revision ID: b4d7e1a9c25f
+Revises: e7f2b3c9d418
+Create Date: 2026-08-15
+
+"""
+
+import sqlalchemy as sa
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "b4d7e1a9c25f"
+down_revision = "e7f2b3c9d418"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    # Se crea nullable para poder rellenar las filas existentes antes de exigir
+    # el NOT NULL: una tabla con datos no admite una columna NOT NULL sin default.
+    op.add_column("quick_matches", sa.Column("play_mode", sa.String(20), nullable=True))
+
+    op.execute(
+        """
+        UPDATE quick_matches
+        SET play_mode = CASE
+            WHEN match_format IS NOT NULL THEN 'SCRATCH'
+            ELSE 'HANDICAP'
+        END
+        WHERE play_mode IS NULL
+        """
+    )
+
+    op.alter_column(
+        "quick_matches",
+        "play_mode",
+        nullable=False,
+        server_default="HANDICAP",
+    )
+
+    op.create_check_constraint(
+        "ck_quick_matches_play_mode",
+        "quick_matches",
+        "play_mode IN ('SCRATCH', 'HANDICAP')",
+    )
+
+
+def downgrade() -> None:
+    op.drop_constraint("ck_quick_matches_play_mode", "quick_matches", type_="check")
+    op.drop_column("quick_matches", "play_mode")
--- a/alembic/versions/b4d7e1a9c25f_add_play_mode_to_quick_matches.py
+++ b/alembic/versions/b4d7e1a9c25f_add_play_mode_to_quick_matches.py
@@ -7,11 +7,16 @@ Las filas existentes NO se rellenan todas igual, y es deliberado: se les asigna
 el modo que reproduce lo que la aplicacion venia mostrando, para no reescribir
 el resultado de partidas ya jugadas.
 
-- Partidas de match play (match_format IS NOT NULL): el resultado se venia
-  calculando con los golpes brutos, sin aplicar handicap. Eso es exactamente
-  SCRATCH, asi que se quedan en SCRATCH y su resultado no se mueve.
-- Partido libre (scoring_format IS NOT NULL): la clasificacion ya se calculaba
-  neta, aplicando el Playing Handicap de cada uno. Eso es HANDICAP.
+- Match play YA TERMINADO o cancelado (match_format IS NOT NULL y status en
+  COMPLETED/CANCELLED): el resultado se venia calculando con los golpes brutos,
+  sin aplicar handicap. Eso es exactamente SCRATCH, asi que se quedan en SCRATCH
+  y su resultado guardado no se mueve.
+- Todo lo demas, HANDICAP. Esto incluye el partido libre (su clasificacion ya se
+  calculaba neta) y, sobre todo, las partidas de match play que en el momento del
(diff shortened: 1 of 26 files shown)


agustinEDev:
@coderabbitai full review

agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai review

PR #213, opened 2026-08-18
feat(scoring): Score every player against their own tee

agustinEDev:
Backend half of the per-tee hole data work. Frontend half is agustinEDev/RyderCupWeb#412 and must deploy right after this.

## The problem

Par, stroke index and distance belong to the **tee**, not to the course. `GolfCourse.holes` was only the card derived from the first tee that had one. Of the 800 imported federated courses with two or more tees:

| Between tees | Courses | |
|---|---|---|
| the metres differ | 574 | 71.8% |
| the stroke index differs | 56 | 7.0% |
| the par differs | 25 | 3.1% |

Stroke allocation already resolved each player's tee — `StrokeContextBuilder` and `TeeContextBuilder` both do. **Everything else did not.** Seven call sites read the reference card:

- Stableford points for a quick match (`get_recent_matches`)
- the round that counts towards a player's average, and its differential (`get_player_stats`, 3 sites)
- the achievement feed, for both quick matches and tournaments (`social`, 2 sites)
- the tournament scoring view (`get_scoring_view`)

So a player not on the first tee was measured against someone else's par, while their strokes were allocated with their own tee's index. The two halves of the same calculation disagreed.

## The fix, and why the rename is the important part

The resolution now lives **once**, in the aggregate: `GolfCourse.tee_for` and `GolfCourse.hole_card_for`, including the gender fallback that both context builders and the frontend had each copied.

**`GolfCourse.holes` is renamed to `reference_card`.** This is the part that stops it happening again. That property is what five different call sites reached for by accident, in two repositories, written at different times — because it was called "the holes" and sat right there. Named for what it actually is, it is no longer the obvious thing to grab. It is derived and never persisted (the mapper rebuilds it from the first tee on load), so no mapping, migration or contract changes.

`countable_holes` now takes the card rather than the course: choosing the card is the caller's job, because only the caller knows who played.

## Contract change

`HoleInfoDTO` gains `meters`, and each `ScoringPlayerDTO` carries its own `hole_card`, resolved server-side so the client cannot resolve it a different way.

`holes` **stays** on the response as the reference card, deprecated in its docstring. Removing it now would break production in the window between deploying this and the frontend — the lesson from the last incompatible change. It goes once the client stops reading it.

## Testing

- `pytest tests/unit/` — 2872 passing
- `mypy src/ --ignore-missing-imports` — clean, 605 files
- `ruff check src/` — clean (the 6 repo-wide warnings are in `main.py` and an integration test, both untouched here)

New tests:

- `GetScoringViewUseCase` **had no unit tests at all**. Four now cover the card per player: par per tee, metres per tee, ordering, and no course loaded.
- `GolfCourse.hole_card_for`: own card, gender fallback, fallback to the reference card, exact gender winning.
- History: a scratch player on the red tee, where holes 1 and 2 play as par 5, firing 4 everywhere.

Both new suites were verified to fail with the bug put back — the history one reports **36 points instead of 38**.

## Deliberately not here

- **The arnés de paridad FE/BE was not regenerated**: it covers stroke allocation, which is unchanged.
- **Pitch & putt** still falls outside the WHS ratings (`MIN_PAR = 66`) — that is #206.
- **Hand-entered courses** have no per-tee metres to serve, because the form never collected them. That is #200 and RyderCupWeb#413.

## Summary by CodeRabbit

* **New Features**
  * Scoring views now show each player’s hole card, including tee-specific pars, stroke indexes, and distances.
  * Hole cards are selected according to tee color, gender, and identifier, with sensible fallbacks.
* **Bug Fixes**
  * Scoring, handicaps, Stableford points, round validity, and achievement detection now use the player’s actual tee card.
  * Course listings and totals consistently use the reference card.
* **Tests**
  * Added coverage for tee-specific scoring, hole cards, achievement calculations, and fallback behavior.

### DIFF
--- a/src/modules/competition/application/dto/scoring_dto.py
+++ b/src/modules/competition/application/dto/scoring_dto.py
@@ -46,14 +46,20 @@ class ScoringPlayerDTO(BaseModel):
     tee_color: str
     playing_handicap: int
     strokes_received: list[int]
+    # El par, el índice y los metros son de la barra de cada jugador, no del
+    # campo: en 56 de los 800 campos federados el índice cambia de una barra a
+    # otra y en 25 el par. Se manda resuelto para que el cliente no tenga que
+    # volver a resolverlo — y no lo resuelva de otra manera.
+    hole_card: list["HoleInfoDTO"] = []
 
 
 class HoleInfoDTO(BaseModel):
-    """Info de un hoyo."""
+    """Info de un hoyo, tal y como se juega desde una salida concreta."""
 
     hole_number: int
     par: int
     stroke_index: int
+    meters: int | None = None
 
 
 class PlayerHoleScoreDTO(BaseModel):
@@ -118,6 +124,10 @@ class ScoringViewResponseDTO(BaseModel):
     team_b_name: str
     marker_assignments: list[MarkerAssignmentResponseDTO]
     players: list[ScoringPlayerDTO]
+    # Tarjeta de referencia del campo (la de la primera salida con tarjeta).
+    # Sirve para saber qué hoyos se juegan; lo que se pinta y con lo que se
+    # puntúa a un jugador es su `hole_card`. Se retira cuando el cliente deje
+    # de leerla.
     holes: list[HoleInfoDTO]
     scores: list[HoleScoreEntryDTO]
     match_standing: MatchStandingDTO
--- a/src/modules/competition/application/services/tee_context_builder.py
+++ b/src/modules/competition/application/services/tee_context_builder.py
@@ -27,7 +27,7 @@ class TeeContext:
 
     tee_ratings: dict[tuple[str, str | None], TeeRating]
     holes_by_stroke_index: list[int]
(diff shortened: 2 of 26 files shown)


agustinEDev:
@coderabbitai full review

agustinEDev:
@coderabbitai Checked this one at runtime before changing anything, and the current code already does what you ask:

```python
>>> course.hole_card_for(None, None) == course.reference_card
True   # and it does not raise
```

`tee_for` scans for a tee whose `color` equals the one passed; with `None` nothing matches, the genderless reserve finds nothing either, and `hole_card_for` falls through to `reference_card`. So a quick match created before the frontend required a tee — the case where `QuickMatchParticipant.tee_color` is `None` — is already scored against the course card rather than crashing.

Your point stands in a different way, though: that behaviour was not pinned by any test. It falls out of no `TeeColor` comparing equal to `None`, not out of a written decision, so a future change to the lookup could lose it silently. Added two assertions covering `(None, None)` and `(None, MALE)` rather than a guard that would duplicate the fallback.

PR #217, opened 2026-08-18
Fix: Count a foursomes hole with one ball per side

### LINKED ISSUE #216: Foursomes scores one ball per side, not one per player
## Problem

A foursomes match scored the way foursomes is actually played — one ball per side, each hole entered by whoever hit it — scores **nothing at all**. Not a wrong result: no result.

`get_quick_match_use_case.py:232` skips any hole that isn't filled in by all four players:

```python
if not all(pid in scores for pid in team_a_ids | team_b_ids):
    continue
```

That condition assumes four balls. FOURBALL has four balls and the condition is right there. FOURSOMES has **one ball per side**, so it can never be satisfied unless both partners duplicate the same score.

## Reproduced on the local cluster (2026-08-18)

Quick match `654c58f1-cf83-43a0-8071-b604d4d16415`, FOURSOMES, Son Parc - Par 71, team A = me + partner:

| Hole | Player | Score |
|---|---|---|
| 1 | me | 5 |
| 2 | partner | 4 |

Team A played two holes in nine strokes. The standing endpoint returns `None` and the UI shows "Todavía no hay hoyos completados". The frontend's own-round total shows **"5 golpes · Del equipo"** — the strokes of whoever is looking at the screen, labelled as the team's (`PersonalRoundCalculator.js:80`, `computeParticipantTotals(me, …)`).

## Proposal

Accept **one score per side per hole** in FOURSOMES:

- The hole counts when **each side** has a score, not when each player does.
- The side's gross total is the sum of its holes, whichever partner entered each one.
- Net score keeps using the team allocation, which is already team-level for foursomes (`stroke_allocation_service.py:148`).

Explicitly out of scope: **FOURBALL keeps four scores per hole.** Two balls, two scores, best ball — the current condition is correct there and must not be generalised away.

No schema change is needed for this: `quick_match_hole_scores` stays keyed by participant, and a side's hole is simply the score that exists. Adding `team` to the score row is a bigger change and can wait until there's a reason.

## Frontend counterpart

RyderCupWeb: one entry box per side in foursomes instead of four, one scorecard per side, and the team total summing the side. Backend ships first — a client sending four scores must keep working while both deploys land.

## Who enters it

**Either partner can enter the side's score — there is just one box per side.** The stroke belongs to the side, not to whoever typed it, so the backend must accept the hole from *either* participant of that side and count it once. This is why the rule is "at least one score per side" rather than "the score of a designated player": nobody should have to hand the phone over.

Authorship is not lost by this: `quick_match_hole_scores.recorded_by_participant_id` already records who entered each score, separately from whose score it is.

agustinEDev:
## What was broken

A foursomes card kept the way the format is actually played — one ball per side, each hole entered by whoever hit it — produced **no result at all**. Not a wrong standing: no standing.

Reproduced on the local cluster before the fix (quick match `654c58f1`, Son Parc - Par 71): hole 1 entered on my box (5), hole 2 on my partner's (4) — nine strokes over two holes — and the classification tab read *"Todavía no hay hoyos completados"*.

## Why

`_compute_standing` skipped any hole not filled in by all four players:

```python
if not all(pid in scores for pid in team_a_ids | team_b_ids):
    continue
```

That is right for FOURBALL, which really does play two balls per side, and impossible in FOURSOMES, which plays one.

## The fix

The hole counts when **each side** has delivered its ball, whichever partner entered it. Fourball is untouched and still requires every player: with a ball missing, the side's best ball can still change.

The assumption lived only in this filter — `ScoringService._best_ball` already ignored the `None`s and, in foursomes, took the single score. The team score lists are now built only from the participants who actually have a score, so a side with one entry no longer raises.

## Tests

Four new cases in `TestFoursomesScoresOneBallPerSide`:

- a hole with one score per side counts, and the standing reads 1UP
- the **partner** may be the one who enters the side's ball
- a hole with only one side entered still does not count
- fourball still requires all four scores

Checked that the two behaviour cases **fail without the fix** (the other two pass either way, which is what makes them useful as guards).

Full unit suite: 2885 passed. `ruff` and `mypy` clean on the touched files.

## Frontend

RyderCupWeb#420 turns the four entry boxes into one per side. It depends on this, and both deploys should go out together: a client sending four scores keeps working with this change, so the order backend-then-frontend is safe.

Closes #216

## Summary by CodeRabbit

* **New Features**
  * Added support for FOURSOMES match scoring, requiring one result per team.
  * Improved net score calculations when participants are absent.
  * Added clearer handling of incomplete holes across match formats.

* **Bug Fixes**
  * Prevented errors when scores are unavailable for absent participants.
  * Ensured FOURBALL matches still require scores from all participants.

### DIFF
--- a/src/modules/quick_match/application/use_cases/get_quick_match_use_case.py
+++ b/src/modules/quick_match/application/use_cases/get_quick_match_use_case.py
@@ -3,6 +3,7 @@
 from decimal import Decimal
 
 from src.modules.competition.domain.services.scoring_service import ScoringService
+from src.modules.competition.domain.value_objects.match_format import MatchFormat
 from src.modules.golf_course.domain.repositories.golf_course_unit_of_work_interface import (
     GolfCourseUnitOfWorkInterface,
 )
@@ -230,7 +231,9 @@ class GetQuickMatchUseCase:
             scores = scores_by_hole.get(hole_number)
             if not scores:
                 continue
-            if not all(pid in scores for pid in team_a_ids | team_b_ids):
+            if not self._hole_is_complete(
+                scores, team_a_ids, team_b_ids, quick_match.match_format
+            ):
                 continue
 
             # `calculate_hole_winner` compara scores NETOS. Pasarle el bruto hacia
@@ -239,10 +242,12 @@ class GetQuickMatchUseCase:
             team_a_scores = [
                 self._net(scores[pid], pid, hole_number, strokes_by_participant)
                 for pid in team_a_ids
+                if pid in scores
             ]
             team_b_scores = [
                 self._net(scores[pid], pid, hole_number, strokes_by_participant)
                 for pid in team_b_ids
+                if pid in scores
             ]
             hole_results.append(
                 self._scoring_service.calculate_hole_winner(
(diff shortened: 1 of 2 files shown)


agustinEDev:
@coderabbitai review

PR #218, opened 2026-08-18
Fix: Score foursomes crossed, as a 1 vs 1

agustinEDev:
## Why

Foursomes plays **one ball per side**, so there are only two cards in the match — and they are kept **crossed, exactly like a 1 vs 1**: each scorer writes down their own side's strokes and marks the rival's.

`ScoringCoverageService` shared the non-scorers out **player by player**, evenly, ignoring sides. That rule is right for a format where everyone plays their own ball. In alternate shot it produced assignments that mean nothing — a scorer carrying one rival but not their own partner.

## The bug this closes

The first attempt gave each scorer the rival pair **and nothing else**, with a safety net for "carry your own ball too when the other side has no scorer". Review showed that net was **dead code for side B** (the creator is always team A and always a scorer) and **switched off for side A** the moment any B player was ticked.

Concrete failure: creator A1 and registered rival B1 both ticked as scorers, B1 never opens the app → **side A's ball cannot be recorded by anyone** and the match ends with no standing. The very failure this work set out to fix.

## What changes

In FOURSOMES, **every scorer covers all four participants**. With a scorer on each side both balls are carried twice; with a single scorer, that one carries both. No ball is ever left without someone able to record it, and no separate fallback rule is needed.

That also restores the invariant the **deployed frontend v2.13.0** relies on (`useQuickMatchScoring.js:87`): a scorer always covers themselves. A backend-first deploy no longer breaks the UI silently.

FOURBALL and SINGLES keep the even, player-by-player split: with a ball each, that is exactly what they need.

## Two scorers writing the same ball

Deliberate, and the same thing that already happens in a 1 vs 1: if the entries disagree, **the pairs sort it out and correct the hole**. Quick matches carry no player/marker validation — that lives in the `competition` module, and it stays there.

RyderCupWeb#421 stores a side's ball under **one canonical participant** (the first player of the side), so in practice the two scorers write the same row and the last entry wins.

## Tests

`TestFoursomesScoresCrossed`:

- with a scorer on each side, each covers all four
- the only scorer carries both balls
- **a scorer covers its own ball even with a rival scorer** — the regression above; fails without this change
- two scorers on the same side plus a rival scorer: exact coverage, not a superset. The previous test claimed this but built a scenario with no rival scorer, so it exercised the fallback branch instead, and its `>=` assertions hid it.

Full unit suite: 2890 passed. `ruff` and `mypy` clean on the touched files.

## Also here

Corrects the `_hole_is_complete` docstring merged in #217: it claimed `_best_ball` "takes the single score in FOURSOMES". It does not — it ignores `match_format` and returns `min()` of whatever it receives. There is a single score because the frontend stores a side's ball under one participant, not because the engine knows the format.

## Related

- RyderCupAm#217 — a hole counts with one ball per side (merged)
- RyderCupWeb#421 / #422 — one box and one card per side, and who keeps the card

## Also fixed here: the history

Found while reviewing this. `GetRecentMatchesUseCase` carried its **own hand-written copy** of the four-score rule and #217 did not touch it, so every finished foursomes quick match would have shown up in "Partidas recientes" **with no result at all** while the same match already read 2UP on its own screen. The rule now lives once, in `hole_completion_service`, and both screens call it.

The entry's own numbers were the other half of the same defect: they came from the scores stored under the reader's participant, so the partner whose name the side's ball happens to be under got a **full 18-hole personal round** — 72 strokes, 36 Stableford points, computed with their individual handicap — while the other partner, in the same match, got nothing. Foursomes has no personal round: both partners now read the **side's gross strokes**, one ball per hole, and no Stableford points. The DTO descriptions that promised points "in any format" say so too.

A side's stroke for a hole is the **lowest** of its rows, which is what `ScoringService._best_ball` uses to decide the hole. There is normally one row — the ball is stored under the first player of the side — but counting one here and awarding the hole with another would leave a match whose strokes do not explain its result.

## And foursomes is not a personal round

`GetPlayerStatsUseCase` had the same asymmetry, with more consequences: in a finished foursomes, the partner holding the ball rows fed a full 18-hole round into the house average **and the WHS differential**, computed with their individual handicap, while the other partner'''s identical round was dropped for having no scores. Two players of one shared alternate-shot ball ended up with different handicap statistics.

Decision taken with the user: **a foursomes round counts for neither**. The pair plays one ball, so neither player played that round whole and it says nothing about how well either of them plays. Tournament foursomes goes out for the same reason. Singles and fourball are untouched — there each player plays their own ball and signs their own round.

## Full check

2898 unit tests pass, `ruff` and `mypy` clean on `src/`. Every behaviour test was checked to fail without its fix.

## Summary by CodeRabbit

* **New Features**
  * Added support for foursomes scoring in recent match history, including shared team scores and results for both partners.
  * Improved completion handling for foursomes and other match formats.
  * Updated scorer coverage for foursomes and fourball matches.
* **Updates**
  * Excluded foursomes from individual player statistics.
  * Clarified foursomes score and Stableford data in match history.
* **Bug Fixes**
  * Corrected results when only one shared team score is recorded.
  * Prevented incomplete fourball scorecards from being treated as complete.

### DIFF
--- a/src/modules/quick_match/application/use_cases/get_quick_match_use_case.py
+++ b/src/modules/quick_match/application/use_cases/get_quick_match_use_case.py
@@ -3,7 +3,6 @@
 from decimal import Decimal
 
 from src.modules.competition.domain.services.scoring_service import ScoringService
-from src.modules.competition.domain.value_objects.match_format import MatchFormat
 from src.modules.golf_course.domain.repositories.golf_course_unit_of_work_interface import (
     GolfCourseUnitOfWorkInterface,
 )
@@ -24,6 +23,9 @@ from src.modules.quick_match.application.services.stroke_context_builder import 
 )
 from src.modules.quick_match.domain.repositories.quick_match_unit_of_work_interface import (
     QuickMatchUnitOfWorkInterface,
+)
+from src.modules.quick_match.domain.services.hole_completion_service import (
+    hole_is_complete,
 )
 from src.modules.quick_match.domain.services.scoring_coverage_service import (
     ScoringCoverageService,
@@ -232,9 +234,7 @@ class GetQuickMatchUseCase:
             scores = scores_by_hole.get(hole_number)
             if not scores:
                 continue
-            if not self._hole_is_complete(
-                scores, team_a_ids, team_b_ids, quick_match.match_format
-            ):
+            if not hole_is_complete(scores, team_a_ids, team_b_ids, quick_match.match_format):
                 continue
 
             # `calculate_hole_winner` compara scores NETOS. Pasarle el bruto hacia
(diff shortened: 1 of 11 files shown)


agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai full review

PR #224, opened 2026-08-20
Fix(history): Leave foursomes tournament cards without Stableford points

### LINKED ISSUE #223: fix(history): Competition foursomes matches are listed with Stableford points
## Description

Foursomes is played with **one ball per side**, so neither partner plays a round of their own. The product rule is that a foursomes match carries **no Stableford points** — the history shows the side's gross strokes and the match result instead.

That rule is applied in the quick match path but not in the competition one, so a tournament foursomes shows up in "Todas tus partidas" with a points figure.

`_to_quick_match_dto` nulls the points out:

```python
stableford_points=None if side_strokes else (
    totals.stableford_points if totals else None
),
```

`_to_competition_match_dto` does not look at the format at all:

```python
stableford_points=points,
```

`src/modules/user/application/use_cases/get_recent_matches_use_case.py:388`

This is backend-only. The web client already renders the match result whenever `stableford_points` is null (`RecentMatchMapper.js:31`, `RecentMatches.jsx:75`), which is exactly why quick match foursomes look right and tournament ones do not.

## Steps to reproduce

1. Create a competition with two 2-player teams and a round with match format FOURSOMES.
2. Generate the match, start it and score a few holes with both sides so they validate.
3. Submit the scorecards so the match completes.
4. Open "Mis estadísticas" for any of the four players and look at "Todas tus partidas".

## Expected behaviour

The foursomes entry shows the match result (`9&7`) and the side's gross strokes, with no Stableford points — the same as a quick match foursomes.

## Actual behaviour

The entry shows **23 pts** for the winning side and **7 pts** for the losing one, alongside the side's 49 / 70 strokes.

## Environment

Local kind cluster, API built from `develop` (2026-08-20), Chrome + Vite dev server.

## Priority

P1 — High

## Additional context

Verified in the browser on 2026-08-20 with match `a6a8ba1a` of "Torneo Par Por Barra" (Son Parc, Par 71). The `/api/v1/users/me/matches` response for all four players:

| player | result | score | stableford_points | total_strokes |
|---|---|---|---|---|
| Agustin / Tercero (side A) | WON | 9&7 | **23** | 49 |
| Rival / Cuarto (side B) | LOST | 9&7 | **7** | 70 |

Player stats are unaffected: `get_player_stats_use_case` already skips FOURSOMES on both the quick match and the tournament path, so the match correctly stays out of the average, the WHS differential and the round count.

agustinEDev:
Closes #223

## What

In foursomes the pair plays **one ball**, so the card is not either partner's own round and carries **no Stableford points** — the history shows the side's gross strokes and the match result instead.

`_to_quick_match_dto` already applies that rule. `_to_competition_match_dto` passed the points straight through, so a tournament foursomes was listed as **23 pts** for the winning side and **7 pts** for the losing one.

```diff
-            stableford_points=points,
+            stableford_points=None if is_foursomes else points,
```

Player stats were never affected: `get_player_stats_use_case` already skips FOURSOMES on both paths, so the match stays out of the average, the WHS differential and the round count. This was only the history line.

## Why the web client needs no change

`RecentMatches.jsx:75` already falls back to the match result when `stableford_points` is null — that is exactly why quick match foursomes looked right and tournament ones did not. Once the API stops sending points, the entry reads `9&7` like its quick match sibling. No coordinated deploy needed.

## Tests

`played_competition_match` can now seed a scorecard per player, which is what makes the new tests meaningful — the existing tournament tests had no card at all, so their `stableford_points is None` said nothing about the format:

- `test_a_scored_tournament_card_carries_its_stableford_points` — nine pars on a par 4 course are 18 points in SINGLES. Fixes the normal case the foursomes rule is carved out of.
- `test_a_foursomes_card_has_no_stableford_points` — the same nine pars in FOURSOMES: no points for either partner, and both read the side's 36 strokes and the same result.

Reverting the one-line fix makes the second test fail with `18 == None`, so it does not bless the bug.

`pytest tests/unit/` 2915 passed · `ruff check` clean on the touched files · `mypy src/` no issues.

## How it was found

Playing a full competition foursomes in the browser against the local cluster on 2026-08-20 (match `a6a8ba1a`, "Torneo Par Por Barra", Son Parc Par 71): the finished match showed up in "Todas tus partidas" with points, two lines above quick match foursomes that correctly showed `9UP`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Recent match history now correctly handles foursomes rounds.
  * Foursomes matches display shared-side stroke and hole totals without individual Stableford points.
  * Scored tournament matches now accurately report strokes, holes played, and Stableford points.

### DIFF
--- a/src/modules/user/application/use_cases/get_recent_matches_use_case.py
+++ b/src/modules/user/application/use_cases/get_recent_matches_use_case.py
@@ -372,6 +372,12 @@ class GetRecentMatchesUseCase:
             else course.reference_card
         )
         total_strokes, holes_played, points = self._scorecard_totals(raw, hole_card)
+        # En foursomes la pareja juega UNA bola, así que la tarjeta no es la
+        # vuelta de ninguno de los dos por separado y no lleva puntos Stableford.
+        # Los golpes del bando sí se enseñan, y son los mismos para los dos
+        # compañeros. Es la misma regla que en partida rápida, donde vive en
+        # `_to_quick_match_dto`: aquí se había quedado sin aplicar.
+        is_foursomes = raw.round_.match_format == MatchFormat.FOURSOMES
 
         return RecentMatchDTO(
             id=str(match.id.value),
@@ -385,7 +391,7 @@ class GetRecentMatchesUseCase:
             tournament_name=raw.tournament_name,
             result=self._result_for_team(match.get_winner(), "A" if in_team_a else "B"),
             score=match.result.get("score") if match.result else None,
-            stableford_points=points,
+            stableford_points=None if is_foursomes else points,
             total_strokes=total_strokes,
             holes_played=holes_played,
             partners=partners,
--- a/tests/unit/modules/user/application/use_cases/test_get_recent_matches_use_case.py
+++ b/tests/unit/modules/user/application/use_cases/test_get_recent_matches_use_case.py
@@ -14,6 +14,7 @@ from uuid import uuid4
 import pytest
 
 from src.modules.competition.domain.entities.competition import Competition
+from src.modules.competition.domain.entities.hole_score import HoleScore
 from src.modules.competition.domain.entities.match import Match
 from src.modules.competition.domain.entities.round import Round
 from src.modules.competition.domain.value_objects.competition_id import CompetitionId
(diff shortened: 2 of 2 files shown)


agustinEDev:
@coderabbitai review

PR #228, opened 2026-08-20
Fix(auth): Derive a name when Google does not send one

### LINKED ISSUE #227: fix(auth): A Google account without a surname is registered nameless and can never log in
## Description

Reported from production on 2026-08-20. Signing in with Google fails with the client-side error **"User entity requires id, email, first_name, and last_name"**, and the account is left permanently unable to log in.

Google only sends `given_name` / `family_name` when the account has a structured name; an account with a single name sends `name` and nothing else. The service substitutes an empty string:

```python
first_name=userinfo.get("given_name", ""),
last_name=userinfo.get("family_name", ""),
```

`src/modules/user/infrastructure/external/google_oauth_service.py:95`

Nothing downstream stops it: `User.create_from_oauth` does not check for an empty name, and `UserResponseDTO` types both as `str`, which Pydantic happily fills with `""`. The API answers 200 with a blank name and the **web client** rejects it at `User.js:26`, because `!first_name` is true for an empty string.

`id` and `email` are ruled out: `_get_user_info` already validates both (`if not google_user_id or not email: raise ValueError`), so the only fields that can arrive empty are the name and the surname.

## Why this is P0 and not a cosmetic failure

The registration is **saved before the response is rendered**: `_auto_register_new_user` commits the user, then the client blows up. That account now exists with a blank name — and on the next attempt `_resolve_user` finds it by its OAuth account and returns the same user with the same blank name. Same error, forever. Retrying never gets out of it, and the user cannot register by email either because the address is taken.

One production account is affected and is being fixed by hand.

## Steps to reproduce

1. Use a Google account whose profile has no surname (`family_name` absent from the userinfo response).
2. Sign in with Google on a fresh account.
3. The error dialog appears; the user row is already in the database with `last_name = ''`.
4. Retry: identical error.

## Expected behaviour

The user signs in. A registration with a name Google did not provide lands on `/auth/complete-profile`, which the `is_new_user` flow already does, and the user fills it in.

## Actual behaviour

The account is created nameless and every subsequent attempt fails with the same client-side error.

## Environment

Production, API 2.10.0. Not a regression from that release: neither `google_oauth_service.py` nor the frontend `User.js` has been touched since Google OAuth was integrated.

## Priority

P0 — Critical

## Additional context

Decided fix: derive the name from `name` when `given_name` / `family_name` are missing and, failing that, from the local part of the email — a real value the user corrects immediately on the complete-profile screen — instead of storing an empty string.

agustinEDev:
Closes #227

## The failure

Signing in with Google died with **"User entity requires id, email, first_name, and last_name"**, reported from production today.

Google sends `given_name` / `family_name` only when the account has a structured name; one with a single name sends `name` and nothing else. The service filled the rest with an empty string, and nothing downstream stopped it — `User.create_from_oauth` does not check for a blank name and `UserResponseDTO` types both as `str`, which Pydantic fills happily with `""`. The API answered 200 with a blank name and the web client rejected it at `User.js:26`, where `!first_name` is true for an empty string.

`id` and `email` were never the problem: `_get_user_info` already validates both, so the only fields that could arrive empty were the name and the surname.

## Why it was P0

The registration is committed **before** the response is rendered, so the account already existed with a blank name when the client blew up. On the next attempt `_resolve_user` finds it by its OAuth account and hands back the same blank name: same error, forever. The user cannot register by email either, because the address is taken. One production account was affected and was fixed by hand.

## The fix

| Google sends | first_name | last_name |
|---|---|---|
| `given_name: Ada`, `family_name: Lovelace` | `Ada` | `Lovelace` |
| `name: Ada Lovelace King` | `Ada` | `Lovelace King` |
| `given_name: Ada`, `name: Ada` (the reported case) | `Ada` | local part of the email |
| nothing | local part of the email | local part of the email |

The email's local part is a real value of the user's rather than an invented surname, and it is corrected immediately: a new registration lands on `/auth/complete-profile` through the existing `is_new_user` flow. No combination can return an empty field — including an address with no local part, which would leave the very hole this closes.

## Tests

`test_exchange_code_handles_missing_optional_fields` asserted `first_name == ""` as the correct outcome and called it "handles missing optional fields". That is what let the defect through, so it now pins the value that keeps the account usable, with a note about why it changed.

Six new tests cover the resolution itself, including the reported single-name case and the invariant that no field ever comes back empty. Reverting the source file alone fails seven of them.

`pytest tests/unit/` 2921 passed · `ruff check` clean · `mypy` no issues.

## Not a regression from v2.10.0

Neither `google_oauth_service.py` nor the frontend `User.js` has been touched since Google OAuth was integrated. The release deployed today is unrelated.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved Google sign-in profile handling when first or last names are missing.
  - Names are now filled using available profile information or the email address, preventing incomplete account details.
  - Added handling for blank and whitespace-only name values.

### DIFF
--- a/src/modules/user/infrastructure/external/google_oauth_service.py
+++ b/src/modules/user/infrastructure/external/google_oauth_service.py
@@ -89,11 +89,13 @@ class GoogleOAuthService(IGoogleOAuthService):
                 if not google_user_id or not email:
                     raise ValueError("Google user info is missing required fields (sub, email)")
 
+                first_name, last_name = self._resolve_names(userinfo, email)
+
                 return GoogleUserInfo(
                     google_user_id=google_user_id,
                     email=email,
-                    first_name=userinfo.get("given_name", ""),
-                    last_name=userinfo.get("family_name", ""),
+                    first_name=first_name,
+                    last_name=last_name,
                     email_verified=bool(userinfo.get("email_verified", False)),
                     picture_url=userinfo.get("picture"),
                 )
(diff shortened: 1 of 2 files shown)


agustinEDev:
@coderabbitai review

PR #233, opened 2026-08-20
Fix(avatar): Accept the HEIC photos an iPhone actually stores

### LINKED ISSUE #232: fix(avatar): A photo picked from the iOS library is rejected as not a valid image
## Description

Reported from production on 2026-08-20, on an iPhone. Picking an existing photo from the iOS library to use as a profile picture fails with **"Ese archivo no es una imagen válida (JPEG, PNG o WEBP)."** Taking a new photo with the camera in the same dialog works.

That difference is the diagnosis: iOS stores library photos as **HEIC**, while a photo captured through the browser is handed over as JPEG. So the upload only fails for the format the iPhone actually saves in — which is every photo the user already has.

## Cause

`ALLOWED_INPUT_FORMATS = {"JPEG", "PNG", "WEBP"}` — `src/modules/user/infrastructure/external/pillow_image_processor.py:16`

and Pillow cannot even open a HEIC to get that far. Verified against the pinned version:

```
Pillow==12.3.0            # requirements.txt:185
HEIC/HEIF extensions registered: none
```

So `Image.open` raises, and the processor answers `InvalidAvatarImageError("El archivo subido no es una imagen válida")`. The API returns 400 and the client turns it into the message above (`useAvatar.jsx:151`).

The `accept="image/jpeg,image/png,image/webp"` on the file input (`AvatarPicker.jsx:127`) does not stop it: on iOS that attribute does not prevent picking a HEIC from the library.

## Steps to reproduce

1. On an iPhone, open Profile → profile picture → **Subir foto**.
2. Choose an existing photo from the library (any photo taken by the phone, i.e. HEIC).
3. The error appears and nothing is uploaded.
4. Take a new photo instead in the same dialog: it uploads fine.

## Expected behaviour

A photo from the phone's library uploads like any other. Everything is normalised to JPEG on the way out anyway — `process_avatar_image` already re-encodes whatever comes in — so accepting HEIC changes nothing downstream.

## Actual behaviour

Every library photo on iOS is rejected, which in practice means an iPhone user can only set an avatar by taking a new picture.

## Environment

Production, API 2.10.1, iOS 26.6.1.

## Additional context — worth checking together

- `pillow-heif` would let Pillow read HEIC with no other change to the flow. Whatever the decision, the **decompression-bomb guard and the size limit must keep applying** to the new format.
- The same validation is used by the profile photo upload (`src/modules/social/application/use_cases/upload_profile_photo_use_case.py`), which is likely to fail exactly the same way from an iPhone — worth reproducing before closing.
- If HEIC is not going to be accepted, the message should at least say so instead of calling the file invalid, and the client could convert it before uploading. Accepting it is the better outcome for the user.

agustinEDev:
Closes #232

## The failure

Picking an existing photo from the iOS library to use as a profile picture came back **"Ese archivo no es una imagen válida (JPEG, PNG o WEBP)"**. Taking a new photo in the same dialog worked.

That difference was the whole diagnosis. An iPhone stores its library in **HEIC**, while a capture taken through the browser is handed over as JPEG — so the upload failed for the only format the phone actually saves in. In practice, the sole way to set an avatar from an iPhone was to take a new picture, never to use one you already had.

Pillow does not read HEIC on its own. Verified against the pinned version: no HEIC/HEIF extension registered, so `Image.open` raised before the format check ever ran and the processor answered `InvalidAvatarImageError`.

## The fix

`pillow-heif` registers the reader and `HEIF` joins `ALLOWED_INPUT_FORMATS`. Nothing downstream changes: `_decode_and_validate` already re-encodes whatever comes in to JPEG, corrects EXIF orientation and guards the pixel ceiling.

Two things found while testing, both pinned by tests because both would break silently:

- **Pillow identifies a HEIC as `HEIF`**, not "HEIC". Listing `"HEIC"` would have done nothing at all while reading as correct in review.
- **pillow-heif consumes the EXIF orientation itself**, handing the image over already rotated with the tag at 1, so `exif_transpose` leaves it alone. There is no double rotation today; if a future version changed that, portrait iPhone photos would come out on their side.

## Scope

`social`'s profile photo upload shares this processor through `process_gallery_image`, so it is fixed too — the issue flagged it as worth checking and it was. Both API descriptions that listed the accepted formats have been updated so the docs stop lying.

## The guards still hold

Accepting HEIC does not loosen anything: a BMP is still rejected, and the pixel ceiling that protects against decompression bombs applies to the new format as well. Both have their own test.

## Verified end to end

The dependency ships its own libheif binaries, so the Dockerfile needed no change — checked by building the image and decoding a HEIC inside the container, which is the real risk with this kind of package.

Then against the local cluster, with that image deployed: uploading a 1200×900 HEIC carrying EXIF orientation 6 returns **201** where it used to return 400, and what comes back out is a **JPEG 512×512 RGB**, identical to any other photo.

`pytest tests/unit/` 2927 passed · `ruff check` clean on the touched files · `mypy src/` no issues. Reverting only the format list fails three of the new tests.

## Summary by CodeRabbit

* **New Features**
  * Added support for uploading HEIC/HEIF images as profile photos and avatars.
  * HEIC images are automatically processed and converted to JPEG for avatars and gallery images.
  * Existing image validation, including format and pixel-size limits, remains in place.
* **Documentation**
  * Updated upload descriptions to list HEIC among supported image formats.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -184,6 +184,12 @@ python-multipart==0.0.32
 # comprimir a JPEG) antes de guardar la foto de avatar subida por el usuario en BD
 Pillow==12.3.0
 
+# pillow-heif - Lector de HEIC/HEIF para Pillow, que no lo trae. Es el formato en
+# el que un iPhone guarda TODAS las fotos de su fototeca, así que sin esto la
+# única forma de poner un avatar desde un iPhone era hacer una foto nueva: la
+# cámara del navegador entrega JPEG y la fototeca entrega HEIC (BE #232)
+pillow-heif==1.5.0
+
 
 # ================================
 # TESTING Y DESARROLLO
--- a/src/modules/social/infrastructure/api/v1/profile_photo_routes.py
+++ b/src/modules/social/infrastructure/api/v1/profile_photo_routes.py
@@ -81,7 +81,7 @@ PHOTO_CACHE_CONTROL = "private, max-age=31536000, immutable"
 @limiter.limit("30/hour")
 async def upload_my_photo(
     request: Request,  # noqa: ARG001 - Required by SlowAPI for rate limiting
-    file: UploadFile = File(..., description="Image file (JPEG, PNG or WEBP)"),
+    file: UploadFile = File(..., description="Image file (HEIC, JPEG, PNG or WEBP)"),
     caption: str | None = Form(None, max_length=280),
     current_user: UserResponseDTO = Depends(get_current_user),
     use_case: UploadProfilePhotoUseCase = Depends(get_upload_profile_photo_use_case),
--- a/src/modules/social/infrastructure/api/v1/profile_photo_routes.py
+++ b/src/modules/social/infrastructure/api/v1/profile_photo_routes.py
@@ -81,7 +81,7 @@ PHOTO_CACHE_CONTROL = "private, max-age=31536000, immutable"
 @limiter.limit("30/hour")
 async def upload_my_photo(
     request: Request,  # noqa: ARG001 - Required by SlowAPI for rate limiting
-    file: UploadFile = File(..., description="Image file (HEIC, JPEG, PNG or WEBP)"),
+    file: UploadFile = File(..., description="Image file (HEIC/HEIF, JPEG, PNG or WEBP)"),
     caption: str | None = Form(None, max_length=280),
     current_user: UserResponseDTO = Depends(get_current_user),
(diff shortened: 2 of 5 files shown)


agustinEDev:
@coderabbitai review

agustinEDev:
Applied in 562e8aa — good catch on the precision: the processor accepts any HEIF, not only Apple's HEIC flavour, so a `.heif` uploads fine and neither description said so. Both now read `HEIC/HEIF`.

PR #236, opened 2026-08-21
feat(quick-match): Allow recording a picked-up hole

agustinEDev:
## What

A player who picks up their ball finishes the hole without a number — on the scorecard, a dash. Quick matches had nowhere to store it: the `score` column was `NOT NULL` and `SubmitHoleScoreBody` demanded an integer, so the frontend's "Raya" button sent `score: null`, Pydantic answered **422**, and the scoring screen showed *"Ese resultado no es válido"*. The competition module already accepted it (`own_score: int | None`), so the same button worked in one screen and not in the other.

Reported from production, with a screenshot of the failing screen.

## The distinction that matters

A picked-up hole is a **recorded** hole, not a pending one. Key present with a null value vs. no row at all — that is what the live standings and being able to finish the match hang on. `compute_participant_totals` now tells them apart by key presence instead of `get(...) is None`, because both cases were `None` and mean the opposite.

## How a dash counts

| Context | Rule |
|---|---|
| Stableford / stroke totals | Net double bogey (`par + 2 + strokes received`), per WHS Rule 3.1 for a hole not holed out. Yields exactly the zero points picking up is worth. |
| Match play | The null propagates to `calculate_hole_winner`, which reads a side with no ball and gives the hole to the opponent. That service already took `list[int \| None]`, so nothing changed there. |
| **Medal** | **Not allowed.** Stroke play requires holing out on every hole, so `QuickMatch.allows_picked_up_holes()` is false and both scoring use cases reject a null score with a **400**. |
| Achievements | Judged as a plain double bogey against par. No achievement comes out of a double bogey, so a legitimate birdie elsewhere in the round can still be shared. |

## Migration

`d3c7a5f18e42` drops the `NOT NULL` on `quick_match_hole_scores.score`. Instant, no table rewrite, and **backward compatible** — the current API never writes a null. The CHECK constraint is untouched: a CHECK over NULL evaluates to NULL and is not considered violated, so real scores are still bounded to 1..15. The downgrade deletes rows with a null score, since they cannot be represented in the old schema.

## Verified

- `pytest tests/unit/` — 2951 passed
- `pytest tests/integration/` — 438 passed, 1 skipped
- `ruff check .` / `mypy src/` — clean (the 6 pre-existing ruff findings on `develop` are untouched)
- **Against the live local cluster**, not only in tests: image rebuilt, migration applied on boot, then a real Stableford match recorded a dash (200, `score: null`, returned as null in the detail) and a Medal match rejected it (400) while still accepting a number (200).

## Deploy order

**FRONTEND FIRST**, and both in a row — the opposite of the usual order here, and worth reading before deploying.

With the backend first, the installed app (v2.14.0) still offers the picked-up button in every format, and its `score: null` submissions would start being **stored**. That version draws a dash exactly like an unscored hole, leaves the hole selector amber forever (`hs.score != null`) and skips the hole in its Stableford while the backend counts it — bad data written by a client that cannot read it back.

With the frontend first there is no window: the old backend keeps answering 422, which is today's behaviour, and the new screen already maps 400 and 422 to the same copy.

## Also in this PR

Competition rounds now count a picked-up hole too (net double bogey), instead of throwing the whole card out of the average and the WHS differential. Without it the same round counted for your average if it was a quick match and not if it was a competition — and that split applied to future rounds, not just old ones. **This recomputes averages and differentials for rounds already played**, which was agreed before making the change.

Frontend: agustinEDev/RyderCupWeb#431

## Summary by CodeRabbit

* **New Features**
  * Added support for recording picked-up holes with an explicit `null` score in eligible quick match formats.
  * Stableford scoring now treats picked-up holes as net double bogeys with zero Stableford points.
  * Match-play results correctly award holes when one player picks up, while halving holes when both do.
  * Picked-up holes can be corrected to numeric scores and vice versa.

* **Bug Fixes**
  * Medal matches now reject picked-up hole submissions.
  * Round statistics and achievements remain accurate when picked-up holes are recorded.

### DIFF
--- a/alembic/versions/d3c7a5f18e42_allow_null_score_in_quick_match_hole_scores.py
+++ b/alembic/versions/d3c7a5f18e42_allow_null_score_in_quick_match_hole_scores.py
@@ -42,11 +42,25 @@ def upgrade() -> None:
 
 
 def downgrade() -> None:
-    # Las rayas no caben en el esquema viejo. Volver atras las BORRA, y el hoyo
-    # pasa de "recogido" a "sin anotar": es la unica lectura que el esquema
-    # anterior sabe representar, y es preferible a que el downgrade reviente
-    # contra el NOT NULL con la tabla ya poblada.
-    op.execute("DELETE FROM quick_match_hole_scores WHERE score IS NULL")
+    # Las rayas no caben en el esquema viejo, y volver atras solo puede
+    # perderlas: un hoyo "recogido" pasaria a "sin anotar", que es lo unico que
+    # el esquema anterior sabe representar. Antes que borrar tarjetas ya
+    # jugadas en silencio, el downgrade se planta y obliga a decidir a mano.
+    #
+    # Sin rayas guardadas —el caso normal, un rollback inmediato— no hay nada
+    # que perder y el downgrade sigue adelante.
+    hay_rayas = (
+        op.get_bind()
+        .execute(sa.text("SELECT EXISTS (SELECT 1 FROM quick_match_hole_scores WHERE score IS NULL)"))
+        .scalar_one()
+    )
+    if hay_rayas:
+        raise RuntimeError(
+            "Cannot downgrade: there are recorded picked-up holes (score IS NULL) "
+            "that the previous schema cannot represent. Export or delete them "
+            "explicitly before rolling back."
+        )
+
     op.alter_column(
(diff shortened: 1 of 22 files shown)


agustinEDev:
@coderabbitai review

agustinEDev:
@coderabbitai review

Issue #239, opened 2026-08-21
feat(user): Add a unique alias to the user profile


agustinEDev:
## Context

Players want to be known in the app by a nickname (their "alias") instead of their legal name: the greeting on the dashboard, the friend search, the scorecards. This issue covers the whole backend side — the field, searching by it, and returning it for display. The frontend is agustinEDev/RyderCupWeb#435.

Product decision: **the alias is unique across all users** (case-insensitive). Two accounts called "Chuchi" would make the alias useless for finding people and would make impersonation trivial.

### Schema + migration

- `users_table` — `src/modules/user/infrastructure/persistence/sqlalchemy/mappers.py:144`
  - `Column("alias", String(20), nullable=True)`
- New Alembic migration, `down_revision = "d3c7a5f18e42"` (current head).
  - Add the column as nullable. No backfill: an empty alias means "use my real name".
  - Partial, case-insensitive unique index so existing NULLs don't collide:
    ```python
    op.create_index(
        "ix_users_alias_lower",
        "users",
        [sa.text("LOWER(alias)")],
        unique=True,
        postgresql_where=sa.text("alias IS NOT NULL"),
    )
    ```
  - `downgrade()` drops the index and the column.

### Domain

- `User` entity — `src/modules/user/domain/entities/user.py`
  - `_alias` attribute + `alias` property, constructor kwarg defaulting to `None`.
  - `display_name` property: `return self._alias or self.get_full_name()`. This is the single source of truth consumed by #241 — nothing else should re-implement the fallback.
  - Extend `update_profile()`, `_validate_profile_update()` and `_detect_profile_changes()` to handle the alias, and include the change in `UserProfileUpdatedEvent`.
  - `get_full_name()` keeps returning the real name, untouched: the RFEG handicap lookup (`update_user_handicap_use_case`, `login_user_use_case`, `generate_matches_use_case`) searches the federation by legal name, and an alias there would silently stop finding handicaps.
- New `AliasValidator` next to `NameValidator` in `src/shared/application/validation/validators.py`:
  - 2–20 characters (the autocomplete needs 2 to trigger, so a 1-char alias would be unfindable).
  - Letters (accents included), digits, space, `.`, `_`, `-`. No emoji, no HTML.
  - Trim, collapse repeated inner spaces, reject leading/trailing spaces after trimming.
  - Reuse `sanitize_html()` first, like the name fields do.
[...]

PR #248, opened 2026-08-31
feat(user): Find people by their alias, and show it next to the name

agustinEDev:
Part 2 of #239: typing somebody's alias finds them. Part 1 shipped in #247; Part 3 (`display_name` in the response DTOs) comes next, on top of `develop`.

## What this adds

**The four queries that search users by name now match the alias too** — the public autocomplete, the admin listing (shared by `find_all`/`count_all`), the competition creator filter, and the in-memory repository so the unit tests exercise the same rule the real one enforces. Without this, setting a nickname made you invisible to the people who only know you by it, which is most of the point.

**The autocomplete returns `alias` next to the real name** (null when there is none). Showing only one of the two leaves you guessing: only the real name and you cannot tell which of the two "Chuchi" you found; only the alias and you cannot tell whose account it is. This does not widen what the search exposes — the alias is public by design — and the email is still absent, with a test saying so.

**The admin listing shows the alias next to the legal name.** It gained alias *matching* here, so without the field a search for "Chuchi" returned rows where nothing visible contained "Chuchi". Next to the real name, never instead of it: #239 is explicit that an admin has to know whose account it is.

## Two things worth reviewing

**A regression I introduced and then fixed** (`8c59065`). The alias branch stripped the query before searching it, and `if search_creator:` is true for `"  "` — so `contains("")` matched every non-NULL alias and `?search_creator=%20` returned the whole first page instead of nothing. The text now goes in as it arrives.

**The competition creator search stops interpolating the query into the pattern.** `ILIKE f"%{x}%"` meant a typed `%` returned every competition and a `_` matched any character. All three branches now use `LOWER(...).contains(..., autoescape=True)`. That part is pre-existing and technically outside #239, but the first version of this branch documented the hazard in a comment while leaving it in place, which was the wrong half of the job.

## Behaviour protected by tests

- A **deactivated account does not reappear** when searched by its alias — it would be a back door into the visibility rule that protects whoever asked to be deactivated. Tested in both repositories.
- **Accounts without an alias are not hidden** by the new OR branch: `LOWER(NULL)` is NULL, so they keep matching through the others. Tested in both repositories, because a search that quietly stopped finding people would be the worst possible way to learn it.
- The creator-alias test seeds a **second competition from a different creator**, so a branch that matched everything fails it.

## Deliberately not done

`find_by_full_name` (behind `GET /users/search?full_name=`) still resolves legal names only, so typing an alias there answers 404. It is an exact-match lookup, it is not one of the four queries #239 lists, and a fallback would mix two namespaces exactly where we accepted that an alias may be another user's real name. Written up on the issue.

## Testing

- 3027 unit tests green; `ruff` and `mypy` clean on everything touched.
- `/code-review high` before pushing; its findings are the last two commits.
- **Integration tests not run locally** — they need Postgres on 5434 and neither Docker nor the Kind cluster was up. CI covers them, and this branch adds four of them.

https://claude.ai/code/session_01PSrmca8GvTCsF7tbjaM7Ss

## Summary by CodeRabbit

* **New Features**
  * User searches and competition creator filters now match aliases alongside names.
  * Search results and administrative listings now include users’ aliases.
  * Autocomplete responses include alias and avatar details while omitting email addresses.
* **Bug Fixes**
  * Deactivated accounts remain excluded from searches.
  * Special characters such as `%` are treated literally rather than as wildcards.
* **Documentation**
  * Updated API and search behavior documentation to reflect alias matching and privacy rules.

### DIFF
--- a/docs/API.md
+++ b/docs/API.md
@@ -295,16 +295,22 @@ Returned in all endpoints that include user data (login, current-user, register,
 **GET /api/v1/users/search-autocomplete** (Authenticated)
 
 **Query Parameters:**
-- `query` (string, required, 2-100 chars) - Partial name to search for
+- `query` (string, required, 2-100 chars) - Partial name or alias to search for
 
 **Response (200 OK):**
 - `users` (array, max 10 results)
   - `user_id` (string, UUID)
-  - `email` (string)
   - `full_name` (string) - "first_name last_name"
+  - `first_name` (string)
+  - `last_name` (string)
+  - `alias` (string or null) - public nickname, if the user set one
+  - `avatar_source`, `avatar_preset_id`, `has_avatar_upload`
 
 **Business Rules:**
-- Case-insensitive partial match on first_name or last_name
+- Case-insensitive partial match on first_name, last_name or alias
+- The alias travels next to the real name, so two similar players can be told apart
+- No email: it was removed so addresses could not be harvested by typing names
+- Deactivated accounts never appear, not even when the query matches their alias
 - Returns max 10 results
 - Only authenticated users can search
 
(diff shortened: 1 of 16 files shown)


agustinEDev:
@coderabbitai full review

PR #249, opened 2026-08-31
feat(user): Ship the resolved display name so the client does not decide

agustinEDev:
Part 3 of #239, and the last of the backend. Parts 1 (#247) and 2 (#248) are already in `develop`. After this, the frontend side (RyderCupWeb#435) has everything it needs.

## What this does

The backend resolves "alias if there is one, full name otherwise" **once**, and ships the answer. `User.display_name` is the only place that fallback lives; without shipping it, every screen in the frontend would carry its own `alias || nombre` and they would drift.

Two shapes, depending on what each DTO already exposed:

- **Where first/last travel separately** (the user's own record, user search, competition creator, enrolled players, player profiles, feed authors) `display_name` is **added**. `first_name` / `last_name` stay, so the frontend in production keeps working and this deploys on its own.
- **Where the field was already a "name to show"** (`user_name`, `scorer_name`, `marks_name`, the quick match participant `name`, friendship and invitation names) no new field was needed: what travels in it changes.

Following the field rather than the issue's list turned up three places it did not mention: the leaderboard, recent matches and competition invitations.

## Decisions worth reviewing

- **A guest in a quick match keeps the typed name.** No account, no alias. That is the `if` of an `if/else` whose `else` did change, so it has its own test — the next edit should not sweep both.
- **Social notification emails use the alias; account emails do not.** "Somebody invited you" / "somebody sent you a request" names that somebody the way the app does. Verification and password reset talk about the account and keep the legal name. Tests on both sides.
- **`display_name` is required, not defaulted.** It is the field the frontend paints with no fallback, so a forgotten one should fail a test rather than render a nameless row. Making it required immediately found three test files building those DTOs without it.
- **An alias can still be somebody else's real name**, and it now travels to more screens and into emails without the real name beside it. That is the trade decided in #239; the only screen that disambiguates on purpose is the autocomplete, which shows alias and full name together.

## The excluded paths, now with tests

The issue said it was not done until each one had a test, and it was right: they all still called `get_full_name()`, but nothing was watching.

The RFEG one matters most. The federation has people on file under their registered name, so if that lookup ever switched to `display_name`, anyone with an alias would stop finding their handicap **silently** — no error, just a "not found" that reads like an RFEG problem. Two tests, both directions.

## Watching, measured

`/code-review` reverted all of `src/` to `develop` and only **2** tests failed — eleven of the thirteen changed places had nothing watching them. The same mutation now fails **5 tests across four files**. Still unwatched, and stated rather than hidden: the leaderboard, the creator and enrolled-player DTOs, the feed, the player profile and the match history. Those need the database.

## Testing

- 3038 unit tests green; `mypy` clean; `ruff` clean on everything touched.
- `/code-review high` before pushing; its findings are the last two commits.
- **Integration tests not run locally** — no Postgres on 5434. CI covers them; this branch adds three.

https://claude.ai/code/session_01PSrmca8GvTCsF7tbjaM7Ss

## Summary by CodeRabbit

- **New Features**
  - Added a resolved display name for users, showing their alias when available and full name otherwise.
  - Display names now appear consistently across profiles, search results, feeds, competitions, invitations, friend requests, leaderboards, scoring views, and Quick Match.
  - Guest participant names remain unchanged.
  - Legal names continue to be used for account-related communications and external identity checks.

- **Tests**
  - Added coverage verifying alias and full-name display behavior across user, social, competition, and Quick Match experiences.


agustinEDev:
@coderabbitai full review

Issue #251, opened 2026-09-01
feat(competition): Stroke play competitions — Stableford or Medal, with a tee sheet


agustinEDev:
Stroke play competitions — Stableford or Medal, individual, with a tee sheet — alongside the Ryder Cup format we have today. Decided with Agustín on 1 Sep 2026; the answers are recorded in each section so nobody re-opens them by accident.

The organiser we are building for is **a club's sports director**, not somebody's mate setting up a weekend. That changes what has to work: forty players, most of them without an account, cards that have to add up, and results that leave the app for whatever the club uses.

## What already exists and must be reused

Worth stating first, because it is more than it looks:

- **Scorer and marker, with agreement.** `MarkerAssignment` records who scores, whom they mark and who marks them, and `SubmitScorecardUseCase` refuses a card whose holes are not `ValidationStatus.MATCH` — the player's numbers and the marker's have to agree. That is exactly the rule a club tournament runs on, and it is built.
- Enrollments, invitations, organiser permissions, golf courses with per-tee cards, hole-by-hole scoring, playing handicap and allowances.
- **`Round` already carries its own `golf_course_id`**, so a different course per round is nearly free.

## What does not exist

- **Everything about a competition is two teams**: `team_1_name`, `team_2_name`, `team_assignment` on the entity, `team_a_players` / `team_b_players` on every `Match`.
- No Stableford or Medal in competitions — those live only in quick match.
- No tee times anywhere in the model.
- No player without an account.
- No enrolling into a **particular round**: today you enroll into the competition and that is that.

## Decisions taken

**1. A type of competition, not a new entity.** "Create competition" asks the type first: Ryder Cup or stroke play. Everything above is reused; what changes is how groups are made and how standings are computed. Teams become something that belongs to the Ryder Cup type.

**2. One modality per competition.** Stableford *or* Medal, chosen once. A club that wants a day of each runs **two competitions** — not a per-round setting.

**3. Players with and without an account.** The organiser adds registered friends and types name and handicap for everybody else. Both kinds play, appear in the draw and appear in the standings.

**4. Categories by handicap range**, defined by the organiser (up to 12, 12 to 26, …).

**A player's category is fixed when the competition starts** and does not move again, even if their handicap does — that is what keeps the tables comparable and stops somebody's accumulated total belonging to two categories at once. The *playing* handicap of each round is still the one in force that day: they are different things.

[...]

PR #255, opened 2026-09-03
feat(user): Let a player choose alias or real name per competition

### LINKED ISSUE #254: feat(user): Let a player choose whether a competition shows their alias or their real name
## Where this stands

Since #239 the alias replaces somebody's name **everywhere**, competitions included: the leaderboard, the scoring view, the enrolled players, the creator of a competition. That is the right default for a quick match among friends — the whole point of an alias is that people see you the way they know you.

A championship is a different setting. A club tournament has a starting sheet, a results list and sometimes a prize table, and there a player may well prefer their real name on it. Right now they have no say: put an alias on, and it shows up in the standings of a tournament too.

## What this asks for

**Quick matches keep the alias, always.** Nothing changes there.

**In competitions the player chooses.** Somebody with an alias decides whether that competition shows them by it or by their legal name.

## To decide before building

- **Per competition, or one setting for all of them?** Choosing when you enroll fits a club tournament and a friendly the same week; a single "in competitions, use my real name" is one decision instead of one per event, and cannot be forgotten. The first is more flexible and more screens.
- **Does the organiser always see the real name?** They probably need it to check who is on the list, whatever the player chose for the public standings. The admin panel already works that way — legal name, alias beside it.
- **What about a card that leaves the app?** Once #415 exists, an exported scorecard is a document. It may want the legal name regardless of the setting, the same way the RFEG lookup and the account emails already do.
- **And somebody who changes their mind mid-tournament?** The standings are read by others between rounds; a name that changes halfway is confusing. Freezing the choice when the competition starts is the same rule already agreed for categories in #251.

## Shape of it

The choice is a piece of data, so it lives in the backend: whatever the player picked has to reach the DTOs of the competition module, which today all resolve `display_name` unconditionally. `User.display_name` stays the single place the alias/name fallback lives — what changes is who asks for it.

The frontend half — where the choice is offered, and what the standings show — gets its own issue in RyderCupWeb once the shape is agreed.

## Related

- #239 — the alias itself, already shipped in v2.13.0
- #251 — stroke play competitions, which will lean on this: a club's starting sheet is exactly the case where the legal name matters

agustinEDev:
Closes #254

## What ships

Since #239 the alias replaces a name everywhere, competitions included. That is right for a quick match among friends; a club tournament has a starting sheet and public standings, and there a player may prefer their legal name.

The player decides, **per enrollment** — not the creator (who already decides the custom handicap), and not once for the whole profile: the same person can play a club tournament under their legal name and a friendly under their alias, the same week. Quick matches are untouched: the alias there is unconditional, always.

`PUT /enrollments/{id}/name-preference`, owner-only. The leaderboard, the scoring view and the enrolled-players list read the player's own enrollment before falling back to the alias.

## Decisions worth naming

**Not frozen when the tournament starts.** The issue floats freezing it (mirroring #251's categories) and we deliberately went the other way: a player who picked wrong should be able to fix it without waiting for the tournament to end. The cost is that a name can change between rounds; that was weighed and accepted.

**The creator's own byline is out of scope.** `CreatorDTO` still reads `display_name` unconditionally. Creators are auto-enrolled, so their preference *is* honoured in the enrolled-players list; what is not is the organiser byline on the competition detail. Threading it there means reaching from the user-facing mapper into competition enrollments, and no screen asks for it yet.

**The name fallback lives in one place.** `User.display_name_for_competition()`, next to `display_name`, whose docstring already claims to be the only home of that fallback.

## Review trail

Three local `/code-review` passes before pushing. What they caught and what was done:

- **Silent truncation past the 100th enrollment.** `find_by_competition` caps at 100 rows, and a competition accumulates rows beyond its player cap over time (rejections, withdrawals, re-enrolments). Fixed with a new repository method, `find_by_user_ids_and_competition`, that asks only for the players actually in the leaderboard — one `IN` query, no offset, nothing to skip. Regression test plants 105 filler enrollments.
- **Name resolution written three times.** Consolidated onto the `User` entity.
- **Two sequential lookups per player in the scoring view.** Now concurrent.

## Verification

- Unit: full `tests/unit/modules/competition` suite green
- Integration: 39 tests green (`test_enrollment_endpoints.py` + the new SQLAlchemy-level repository tests), run isolated — combined with the unit suite the local box runs out of connections creating per-worker databases, which also breaks pre-existing files it does not touch
- `ruff` and `mypy` clean on everything touched; single Alembic head; app boots with the route registered
- Each behaviour checked by mutation

https://claude.ai/code/session_01Rze3QE1CDwfjmRbivTrbui

## Summary by CodeRabbit

* **New Features**
  * Players can choose whether their alias or real name appears for each competition.
  * Added an endpoint for updating this preference, with ownership validation.
  * Leaderboards, scoring views, and enrollment lists now display names according to each player’s selection.
  * The preference can be changed throughout the competition lifecycle.

* **Bug Fixes**
  * Name preferences are now correctly applied per competition and enrollment.

### DIFF
--- a/alembic/versions/4aeb1cb3223f_add_name_preference_to_enrollments.py
+++ b/alembic/versions/4aeb1cb3223f_add_name_preference_to_enrollments.py
@@ -1,0 +1,42 @@
+"""add name preference to enrollments
+
+Desde #239 el alias sustituye el nombre en todas partes, competiciones
+incluidas: la clasificación, la anotación, la lista de inscritos. Es lo
+correcto en una partida entre amigos, pero un torneo de club tiene lista de
+salida y clasificación pública, y ahí un jugador puede preferir su nombre
+legal (BE #254).
+
+`use_real_name` es por INSCRIPCIÓN, no del perfil ni una sola vez por
+usuario: la misma persona puede jugar un torneo de club con su nombre legal
+y una liguilla de amigos con su alias, la misma semana.
+
+`false` por defecto — se sigue enseñando el alias, que es el comportamiento
+de antes de esta migración — y NOT NULL: no hay un tercer estado «sin
+decidir» que resolver en cada lectura.
+
+Revision ID: 4aeb1cb3223f
+Revises: c8e1d5b73f92
+Create Date: 2026-09-03
+
+"""
+
+import sqlalchemy as sa
+
+from alembic import op
+
+# revision identifiers, used by Alembic.
+revision = "4aeb1cb3223f"
+down_revision = "c8e1d5b73f92"
+branch_labels = None
+depends_on = None
+
+
+def upgrade() -> None:
+    op.add_column(
+        "enrollments",
+        sa.Column("use_real_name", sa.Boolean(), nullable=False, server_default=sa.false()),
+    )
+
+
+def downgrade() -> None:
+    op.drop_column("enrollments", "use_real_name")
--- a/src/config/dependencies.py
+++ b/src/config/dependencies.py
@@ -128,6 +128,9 @@ from src.modules.competition.application.use_cases.send_invitation_by_user_id_us
 from src.modules.competition.application.use_cases.set_custom_handicap_use_case import (
     SetCustomHandicapUseCase,
 )
+from src.modules.competition.application.use_cases.set_name_preference_use_case import (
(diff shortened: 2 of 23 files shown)


agustinEDev:
@coderabbitai full review

PR #272, opened 2026-09-06
perf(auth): Run bcrypt outside the event loop

agustinEDev:
## What

bcrypt with 12 rounds costs ~200ms of CPU, and every call sat directly inside a coroutine. Production runs a **single uvicorn worker**, so while a password was hashed the process served no other request: one login froze the whole API, and the password-history check (up to 5 verifications in a row) froze it for over a second.

Measured against production before the change:

| Request | Time |
|---|---|
| `/health` (no DB) | ~86 ms |
| `POST /auth/login` (401) | ~589 ms |

## How

All **seven** hashing call sites now go through `run_bcrypt()`: login, change password (current-password check, the history loop and the new hash), password reset, registration and system-user bootstrap.

`run_bcrypt()` has two halves, and the second matters as much as the first:

1. **It hands the call to a thread**, freeing the event loop. bcrypt releases the GIL while it works, so the loop keeps dispatching everything else.
2. **It uses its own small pool** (`BCRYPT_MAX_CONCURRENCY`, default 2). Without that cap, unblocking the loop would trade a latency problem for a resource-exhaustion one (OWASP A04): N simultaneous logins would fire N hashes at half a core. Requests above the cap wait on the loop, which is cheap.

The context is copied into the thread the same way `asyncio.to_thread` does, so the correlation-id keeps showing up in logs written from inside the hash.

`env_int()` reads these integers with a safe fallback and a warning, so a mistyped or blank variable cannot take the container down at import time. `DB_POOL_SIZE` / `DB_MAX_OVERFLOW` make explicit the pool sizes that were implicit in SQLAlchemy's defaults — **behaviour is unchanged with a single process**.

## Architecture

No new dependency direction: the domain stays untouched, and 9 other use cases already import from `shared.infrastructure`. `user.py` has no `await` and no I/O, so running its methods in a thread is safe.

## Not in this PR

Found while reviewing; each deserves its own issue:

- **Login does not dummy-hash for unknown emails**, so response time tells registered addresses apart from unregistered ones (pre-existing, OWASP A07).
- **`failed_login_attempts` is written as an absolute value with no row lock**, so parallel attempts can lose increments and soften the lockout (pre-existing).
- **Registration and password change hash inside an open transaction**, holding a DB connection while they wait.
- **slowapi keeps rate limits in per-process memory**: raising `WEB_CONCURRENCY` — which uvicorn reads on its own, no repo change needed — would multiply every rate limit by the worker count. Needs shared storage before that knob is turned up.

## Testing

- 3105 unit tests pass, `mypy` clean, `ruff` clean on the touched files.
- New tests include the two that actually guard the fix: one fails if the thread hop is removed, another if the concurrency cap is.

## Summary by CodeRabbit

- **Security**
  - Improved password creation, verification, and reset processing while preserving existing authentication and account-protection behavior.
  - System-generated passwords are now stored using secure one-way hashing.

- **Performance**
  - Password-related operations no longer block other application activity, improving responsiveness during authentication and account updates.

- **Configuration**
  - Database connection pool capacity can now be configured through environment settings, with safe defaults and validation.

### DIFF
--- a/src/config/database.py
+++ b/src/config/database.py
@@ -8,6 +8,8 @@ from sqlalchemy.ext.asyncio import (
     create_async_engine,
 )
 
+from src.shared.infrastructure.env import env_int
+
 # Cargar las variables de entorno del fichero .env
 load_dotenv()
 
@@ -36,6 +38,13 @@ async_engine = create_async_engine(
     echo=False,
     pool_pre_ping=True,  # Detecta conexiones muertas antes de reutilizarlas
     pool_recycle=1800,  # Recicla conexiones cada 30 min (evita stale connections)
+    # El pool es POR PROCESO, y uvicorn lee WEB_CONCURRENCY del entorno por su cuenta
+    # (uvicorn/config.py), sin que este repo la declare en ninguna parte: subirla
+    # multiplica las conexiones contra Postgres por N x (pool_size + max_overflow), que
+    # los planes pequeños no dan. Los valores por defecto son los de SQLAlchemy, así que
+    # mientras haya un solo proceso nada cambia.
+    pool_size=env_int("DB_POOL_SIZE", default=5),
+    max_overflow=env_int("DB_MAX_OVERFLOW", default=10),
 )
 
 
--- a/src/modules/user/application/use_cases/ensure_system_user_use_case.py
+++ b/src/modules/user/application/use_cases/ensure_system_user_use_case.py
@@ -12,6 +12,7 @@ from src.modules.user.domain.repositories.user_unit_of_work_interface import (
 )
 from src.modules.user.domain.value_objects.email import Email
 from src.modules.user.domain.value_objects.user_id import UserId
+from src.shared.infrastructure.security.bcrypt_executor import run_bcrypt
 
 # Longitud holgada: nadie va a teclear esta contraseña nunca, y cuanto más
 # larga, menos margen deja si algún día la cuenta se reactivara por error.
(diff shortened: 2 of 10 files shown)


agustinEDev:
Follow-ups from the review, now tracked (referenced, not closed by this PR):

- #281 — login does not dummy-hash unknown emails (user enumeration by timing)
- #282 — `failed_login_attempts` can lose increments, softening account lockout
- #283 — password hashing happens inside open transactions

The fourth finding, rate limits living in per-process memory, is covered by #273 as part of the scaling work.

agustinEDev:
@coderabbitai full review

agustinEDev:
Good catch — fixed in dc7ec70.

An `asyncio.Semaphore` the same size as the pool now sits in front of the executor. With exactly as many permits as threads, nothing reaches the executor without a free thread waiting for it: the internal queue stays empty, and callers wait on the event loop instead — which costs nothing and, unlike the queue, holds no arguments.

One implementation note: the semaphore is per event loop, kept in a `WeakKeyDictionary`. `asyncio.Semaphore` binds to the loop it is first used on, so a single module-level one breaks as soon as there is more than one loop — which is exactly what pytest-asyncio does, one per test.

Added `test_does_not_let_work_pile_up_in_the_executor_queue` as the saturation test you asked for. Verified by mutation: with the semaphore removed, 18 tasks pile up inside the executor and the test fails.

On the suggestion to reject rather than wait — I did not take that part. Returning 503 on login changes the API contract and, badly calibrated, turns away legitimate users. Process-wide backpressure (uvicorn's `--limit-concurrency`) is the right tool for that and protects every endpoint rather than only bcrypt; it is tracked with the rest of the scaling work in #279.

PR #285, opened 2026-09-07
perf(db): Index golf_course_tees by golf_course_id

### LINKED ISSUE #284: golf_course_tees is missing a plain index on golf_course_id (36M tuples scanned)
## The finding

`golf_course_tees` accounts for the overwhelming majority of this database's I/O. From `pg_stat_user_tables` in production:

| Table | seq_scan | seq_tup_read | idx_scan | rows |
|---|---|---|---|---|
| **golf_course_tees** | **31,071** | **35,997,275** | 85,750 | 4,321 |
| users | 94,245 | 1,398,294 | 1,033 | 21 |
| golf_course_tee_holes | 15 | 700,812 | 92,987 | 77,922 |
| golf_courses | 889 | 532,442 | 33,533 | 804 |

**36 million tuples read sequentially** from a 4,321-row table — an order of magnitude more than everything else combined. For contrast, `golf_course_tee_holes` is 18x larger but is read almost entirely through its index (`ix_tee_holes_tee_id`), which is what a healthy table looks like here.

## Why it happens

The table has three indexes, and none of them can serve a lookup by course:

```
golf_course_tees_pkey                     btree (id)
uq_golf_course_tees_color_gender          btree (golf_course_id, color, COALESCE(tee_gender, 'NONE'))
                                          WHERE (color <> 'OTHER')
uq_golf_course_tees_identifier_gender     btree (golf_course_id, identifier, COALESCE(tee_gender, 'NONE'))
                                          WHERE (color = 'OTHER')
```

Both composite indexes lead with `golf_course_id`, but both are **partial**. PostgreSQL only uses a partial index when it can prove the query implies the index predicate, and a plain `WHERE golf_course_id = ?` implies neither `color <> 'OTHER'` nor `color = 'OTHER'`. So the planner has nothing usable and scans the table.

The two partial indexes exist to enforce uniqueness rules, and they should stay as they are — this is not a case of fixing them, but of adding the one that is missing.

## Fix

```sql
CREATE INDEX CONCURRENTLY ix_golf_course_tees_golf_course_id
    ON golf_course_tees (golf_course_id);
```

In an Alembic migration, `CONCURRENTLY` so it does not lock the table.

## Note on scale

At 23 MB the whole database sits in RAM, so this costs little wall-clock time **today**. It is worth doing anyway because it is one line, it is the single largest source of I/O in the system, and tee lookups happen whenever a scorecard is drawn — the most frequent read path in the app.

## Acceptance

- `EXPLAIN ANALYZE` on a lookup by `golf_course_id` shows an index scan.
- `seq_tup_read` on this table stops growing after the migration (compare over a few days; `pg_stat_user_tables` counters are cumulative).

agustinEDev:
Closes #284

## What

`golf_course_tees` accounts for most of this database's I/O. Measured in production on 7 Sep 2026:

| Table | seq_scan | seq_tup_read | idx_scan | rows |
|---|---|---|---|---|
| **golf_course_tees** | **31,071** | **35,997,275** | 85,750 | 4,321 |
| users | 94,245 | 1,398,294 | 1,033 | 21 |
| golf_course_tee_holes | 15 | 700,812 | 92,987 | 77,922 |

36 million tuples read sequentially from a 4,321-row table — more than everything else combined. `golf_course_tee_holes` is 18x larger and reads through its index, which is what healthy looks like here.

## Why it happens

Not visible in the models: the table already has two indexes leading with `golf_course_id`, but both are **partial** (`WHERE color <> 'OTHER'` and `WHERE color = 'OTHER'`). PostgreSQL only uses a partial index when the query implies its predicate, and a bare `WHERE golf_course_id = ?` implies neither. The planner has nothing usable.

Proved rather than assumed — with `enable_seqscan=off`, so a sequential scan costs 10 billion:

```
BEFORE:  Seq Scan on golf_course_tees  (cost=10000000000.00..10000000015.55)
AFTER:   Bitmap Index Scan on ix_golf_course_tees_golf_course_id  (cost=0.00..4.16)
```

It preferred a 10-billion-cost scan over both existing indexes. The two partial ones stay as they are — they do their uniqueness job; what was missing is the plain one.

## From the review

- **The migration drops a leftover invalid index before creating.** A `CREATE INDEX CONCURRENTLY` that dies half-way leaves `indisvalid = false`, which the planner ignores. `entrypoint.sh` retries `alembic upgrade head` on every container start, so a plain `IF NOT EXISTS` would see the name taken, skip, and let Alembic stamp the revision — **reporting the fix as deployed while the scans continue**. Verified both ways: with `IF NOT EXISTS` alone the index stays invalid (`already exists, skipping`); with the drop it recovers to `indisvalid=true`.
- **`env.py` now sets `transaction_per_migration=True`**, which `autocommit_block` requires. Without it one run wraps every pending migration in a single transaction, and this one would commit all of them on its way past — a later failure would leave the database half-migrated.
- **The index is declared in the mapper too**, unlike `ix_tee_holes_tee_id`. It is a plain single-column index that loses nothing in translation, so the schema tests build from metadata matches production.

## Testing

Against a disposable Postgres — deliberately not the local Kind, where applying a branch migration puts the API in CrashLoopBackOff:

- 45 migrations from scratch, with `transaction_per_migration` active
- Index created, `indisvalid=true`
- Invalid-index recovery, and the contra-proof that the previous version left it broken
- Downgrade clean, 0 indexes left
- 3105 unit tests, mypy and ruff clean

## Also here

`test_does_not_block_the_event_loop_while_hashing` (added yesterday in #272) was flaky by design: it required 3 ticker rounds of 10ms inside a 200ms window, which measures machine load as much as it measures the fix. It failed at load average 39 with nothing broken, and passed 5/5 alone. It now compares the counter before and after — with the loop blocked the ticker does not advance at all, so one advance separates the cases. Mutation-checked: removing the thread hop still fails it.

Unrelated to the index, but it would have made this PR's CI unreliable.

## Summary by CodeRabbit

- **Performance**
  - Added an index to improve the speed and responsiveness of golf course tee searches and related data retrieval.

- **Reliability**
  - Database updates now complete in separate steps, allowing interrupted updates to resume more safely and reducing disruption during deployment.

- **Bug Fixes**
  - Improved handling of database index creation so updates can recover cleanly from incomplete operations.

### DIFF
--- a/alembic/env.py
+++ b/alembic/env.py
@@ -111,7 +111,19 @@ def run_migrations_online() -> None:
     )
 
     with connectable.connect() as connection:
-        context.configure(connection=connection, target_metadata=target_metadata)
+        context.configure(
+            connection=connection,
+            target_metadata=target_metadata,
+            # Una transacción POR migración, no una para todas las pendientes. Lo pide
+            # `autocommit_block()`, que usa la migración del índice de golf_course_tees
+            # (43bc18ca6df1): CREATE INDEX CONCURRENTLY no puede ir dentro de una
+            # transacción, así que el bloque cierra la que esté abierta. Sin esto, esa
+            # migración confirmaría de paso todas las anteriores del mismo arranque, y un
+            # fallo posterior dejaría la base a medias en vez de deshacerse entera.
+            # A cambio, cada revisión queda sellada al terminar: un reintento sigue por
+            # donde se quedó en lugar de repetirlas todas.
+            transaction_per_migration=True,
+        )
 
         with context.begin_transaction():
             context.run_migrations()
--- a/alembic/versions/43bc18ca6df1_add_index_on_golf_course_tees_golf_.py
+++ b/alembic/versions/43bc18ca6df1_add_index_on_golf_course_tees_golf_.py
@@ -1,0 +1,75 @@
+"""add index on golf_course_tees.golf_course_id
+
+`golf_course_tees` se lleva la mayor parte del I/O de la base de datos: medido en
+producción el 7 sep 2026, **36 millones de tuplas leídas secuencialmente** en 31.071
+escaneos completos, sobre una tabla de 4.321 filas. Para comparar, `golf_course_tee_holes`
+es 18 veces más grande y se lee casi siempre por índice.
+
+La causa no se ve leyendo los modelos: la tabla YA tiene dos índices que empiezan por
+`golf_course_id`, pero los dos son **parciales**
+
+    uq_golf_course_tees_color_gender       ... WHERE (color <> 'OTHER')
+    uq_golf_course_tees_identifier_gender  ... WHERE (color = 'OTHER')
+
+y Postgres solo usa un índice parcial cuando puede demostrar que la consulta implica su
+predicado. Un `WHERE golf_course_id = ?` no implica ninguno de los dos, así que el
+planificador se queda sin índice utilizable y recorre la tabla entera. Los dos parciales
+se quedan como están: cumplen su función de unicidad, y lo que falta es el simple.
+
+CONCURRENTLY para no bloquear escrituras mientras se construye, lo que obliga a salir de
+la transacción de Alembic (`autocommit_block`).
+
+**Por qué se borra antes de crear.** Un CREATE INDEX CONCURRENTLY que muere a medias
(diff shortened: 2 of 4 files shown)


agustinEDev:
@coderabbitai review

PR #290, opened 2026-09-07
chore(deps): Keep audit tooling out of the production image

### LINKED ISSUE #289: Audit tooling ships inside the production image, and drags nltk with it
## Problem

`safety` and `pip-audit` are declared in `requirements.txt`, which is what `docker/Dockerfile:28` installs into the production image. They are audit tools: nothing inside the running API imports them, and the CI job that uses them installs `requirements-dev.txt` instead (`ci_cd_pipeline.yml`, *Security Checks* → `pip install -r requirements-dev.txt`).

They do not travel alone. Verified with `pip show` in the project venv:

| Package | Required-by | Pinned in |
|---|---|---|
| `nltk==3.10.3` | `safety` | requirements.txt:161 |
| `filelock==3.20.3` | `safety` | requirements.txt:107 |
| `msgpack==1.2.1` | `CacheControl` ← `pip_audit` | requirements.txt:169 |

So the production image carries a natural-language toolkit that no code path touches. `grep -rn "import nltk|from nltk"` over `src/`, `main.py` and `scripts/` returns nothing.

## Why now

Dependabot has two open high alerts, #39 and #48 — the same advisory counted twice, once per manifest:

- `GHSA-8mgp-746c-j5xp` / `CVE-2026-81726`: *NLTK: Model-artifact APIs bypass pathsec and touch files outside allowed roots*, affecting `nltk <= 3.10.3`.
- **There is no patched version.** `first_patched_version` is empty and 3.10.3 is already the latest — the one pinned on 4 Sep to clear the previous five CVEs.

There is no exploitation path here today: the vulnerable APIs have to be called, and nothing calls them. But the package cannot be fixed by upgrading, so the only way to stop carrying the risk is to stop shipping it.

This is also why CI stays green while Dependabot complains: `pip-audit` only fails the build for CVEs that have a fix available (`fix_versions | length > 0`, `ci_cd_pipeline.yml:390`). Both tools see the same advisory; the pipeline deliberately does not block on something that cannot be fixed.

## Proposal

Move to `requirements-dev.txt`, keeping the pins and their comments:

- `safety==3.7.0` — replacing the looser `safety>=3.2.0` already declared there, which currently conflicts with the production pin
- `pip-audit==2.10.0`
- `nltk==3.10.3`, `filelock==3.20.3`, `msgpack==1.2.1` — pinned only because those two tools drag them in

`requirements-dev.txt` starts with `-r requirements.txt`, so CI keeps resolving exactly the same set and the *Security Checks* job is unaffected.

**Staying in `requirements.txt`**, because production genuinely needs them:

- `click==8.3.3` — required by `uvicorn` (and `typer`), not only by `nltk`/`safety`
- `urllib3==2.7.0` — required by `requests` and `sentry-sdk`
- `mako==1.3.12` — required by `alembic`

## Acceptance

- The production image no longer installs `nltk`, `safety` or `pip-audit`.
- The *Security Checks* job still runs both scanners and reports the same results.
- Dependabot alert #39 (the `requirements.txt` one) closes on its own; #48 stays, correctly, since the tooling really does live in dev.

agustinEDev:
Closes #289

## What

`safety` and `pip-audit` were declared in `requirements.txt` — the file `docker/Dockerfile:28` installs into the production image. Nothing in the running API imports them, and the CI job that uses them installs `requirements-dev.txt` (which starts with `-r requirements.txt`, so it keeps resolving the same set).

They do not travel alone. `nltk`, `authlib`, `marshmallow`, `filelock` and `msgpack` are pinned **only** because those two tools pull them in — the comments in the file said so already. That is how a natural-language toolkit ended up shipping to production, and why Dependabot has two open high alerts for `CVE-2026-81726`, an nltk advisory **with no patched version**: it cannot be closed by upgrading, only by not shipping the package.

`zipp` goes too. Its stated consumer, `importlib-metadata`, only ever reached the tree through the audit tooling, so this change is what turns that pin into an orphan.

## What stays, and why

| Package | Real consumer in production |
|---|---|
| `click` | `uvicorn` |
| `urllib3` | `requests`, `sentry-sdk` |
| `mako` | `alembic` |
| `pygments` | `pytest`, which is still declared in `requirements.txt` |
| `setuptools` | nobody declares it — kept on purpose (see below) |

`pygments` is the one the verification caught: I had moved it, and the dry-run showed it still resolving into production, now unpinned at 2.21.0. `setuptools` stays because the base Python image ships it anyway and libraries reach for `pkg_resources` at runtime without declaring it, so the pin is what guarantees the version present has no known CVEs. Both reasons are now written next to the pins.

## Verification

`pip install --dry-run --ignore-installed` against both files, before and after:

| | Before | After |
|---|---|---|
| Production packages | 100 | **62** |

- **No new package** and **no version change** among those that stay.
- `requirements-dev.txt` still resolves all 135, same versions.
- None of the 38 removed modules is imported under `src/`, `main.py`, `alembic/` or `scripts/`. The single grep hit was `import yaml` in `logging/factory.py`, which is PyYAML — still in production — not `ruamel.yaml`.
- Nothing under `docker/`, `entrypoint.sh`, `k8s/` or `scripts/` invokes `safety`, `pip-audit` or `nltk`.

## What this does NOT do

**It does not take nltk off the CI gate.** The audit steps run `pip-audit` and `safety scan` against the *installed environment*, not against a requirements file, and this branch installs `requirements-dev.txt` there — nltk included. The next nltk CVE that ships *with* a fix version will fail the pipeline exactly as before. Changing that means pointing the scanners at `requirements.txt` explicitly, which narrows what gets audited, so it is a separate decision. Noted in the file so the next reader does not assume otherwise.

**It leaves a bigger instance of the same problem.** `pytest`, `pytest-xdist`, `pytest-cov`, `pytest-asyncio` and `pytest-json-report` are still in `requirements.txt`, so the production image also ships the whole test suite. That is a strictly larger dev-only surface than the five packages moved here, but it deserves its own issue — in particular, checking first whether any k8s job runs tests inside the container.

## Summary by CodeRabbit

* **Chores**
  * Added pinned security-auditing tools for development and maintenance workflows.
  * Updated development setup to use consistent versions for security checks.
  * Removed security-auditing tools and their related version pins from the application’s runtime requirements.
  * Retained existing runtime version constraints and clarified their accompanying documentation.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -37,9 +37,69 @@ sqlalchemy[mypy]>=2.0.31
 # Detecta: SQL injection, uso de eval(), contraseñas hardcodeadas, etc.
 bandit[toml]>=1.7.9
 
-# Safety - Verifica vulnerabilidades conocidas en dependencias
-# Usa la base de datos de PyUp.io para detectar CVEs
-safety>=3.2.0
+# safety - Auditoría de vulnerabilidades en dependencias
+# Escanea el proyecto contra Safety DB (40K+ vulnerabilidades conocidas)
+# Detecta CVEs en dependencias directas y transitivas
+# Cumple con OWASP Top 10: A06 (Vulnerable and Outdated Components)
+safety==3.7.0
+
+# pip-audit - Auditoría oficial de PyPA para dependencias
+# Escanea contra PyPI Advisory Database y OSV (Open Source Vulnerabilities)
+# Alternativa oficial mantenida por Python Packaging Authority
+# Cumple con OWASP Top 10: A06 (Vulnerable and Outdated Components)
+pip-audit==2.10.0
+
+# ================================
+# PINS TRANSITIVOS DE LAS HERRAMIENTAS DE AUDITORÍA
+# ================================
+# Ninguno de estos lo usa la aplicación: los arrastran safety y pip-audit, que
+# solo se ejecutan aquí y en el job "Security Checks" del CI. Viven en este
+# fichero para que la imagen de producción no los instale (#289).
+# Comprobado con `pip show <paquete>`: su "Required-by" es exclusivamente de dev.
+# Los que sí necesita producción se quedan en requirements.txt — click (uvicorn),
+# urllib3 (requests, sentry-sdk) y mako (alembic).
+
+# filelock - Sincronización de archivos mediante locks (dependencia transitiva de virtualenv/safety)
+# Fijada explícitamente para resolver CVE-2025-68146 (TOCTOU race condition) y CVE-2026-22701
+filelock==3.20.3
+
+# authlib - OAuth and OpenID Connect library (dependencia transitiva de safety)
+# NO la usa el login con Google del proyecto: nada en src/ la importa.
+# Fijada explícitamente para resolver:
+# - CVE-2025-61920 (DoS via large base64 tokens)
+# - CVE-2025-62706 (DoS via zip bomb decompression)
+# - CVE-2025-68158 (Login CSRF vulnerability - fixed in 1.6.6)
+# - CRITICAL (CVSS 9.3): Improper Verification of Cryptographic Signature (x2, fixed in 1.6.7 + 1.6.9)
+# - HIGH (CVSS 8.7): Not Failing Securely - Failing Open (fixed in 1.6.9)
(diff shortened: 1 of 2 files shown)


agustinEDev:
@coderabbitai review

PR #292, opened 2026-09-08
chore(deps): Keep the test runner out of the production image

### LINKED ISSUE #291: The production image installs pytest and its plugins, with no tests to run
Follow-up to #289, same class of problem and a bigger instance of it.

## Problem

`docker/Dockerfile:27-28` installs only `requirements.txt` into the production image, and that file declares the test runner:

```
pytest==9.0.3
pytest-json-report==1.5.0
pytest-xdist==3.8.0
pytest-asyncio==1.4.0
pytest-cov
```

The image cannot even use them: `.dockerignore:32` excludes `tests/`, so it ships a test runner and nothing to run.

Measured with `pip install --dry-run --ignore-installed -r requirements.txt`, that is **10 of the 62 packages** currently in the image:

`pytest`, `pytest-asyncio`, `pytest-cov`, `pytest-xdist`, `pytest-json-report`, `pytest-metadata`, `pluggy`, `iniconfig`, `execnet`, `coverage`.

## Why it matters

Same argument as #289: scanners look at what is *installed*, not at what runs. Dependabot reads the manifests and Trivy scans the built image, so a CVE in `pytest` or `coverage` surfaces as a production alert for code that never executes there — and if it ships without a patched version, it cannot be closed at all. That is exactly what `CVE-2026-81726` did with `nltk`.

No functional risk: nothing breaks today, and no test runs in production. This is attack surface that does not need to be carried.

## Checked before proposing

Nothing runs pytest inside the container — not `entrypoint.sh`, not `docker/`, not `k8s/`, not the `Makefile`. CI runs the suite directly on the GitHub runner (`ci_cd_pipeline.yml:129`, `:212`, `:282`), never in the image.

## Proposal

Move the whole testing block to `requirements-dev.txt`, keeping pins and comments.

`pygments==2.20.0` goes with it: it is pinned in `requirements.txt` **only** because pytest pulls it in, and the comment added in #289 says so explicitly.

Expected result: **62 → 52 packages** in the production image.

## Acceptance

- The production image no longer installs pytest, its plugins or their tree.
- `requirements-dev.txt` resolves to the same set it does today, with the same versions.
- The full CI suite still runs and passes.

agustinEDev:
Closes #291 · Follow-up to #289, same class of problem and a bigger instance of it.

## What

`docker/Dockerfile:27-28` installs only `requirements.txt`, and that file declared `pytest` and four of its plugins. The image could not even use them: `.dockerignore:32` excludes `tests/`, so it shipped a test runner and nothing to run.

`pygments` moves with them. It was pinned in `requirements.txt` **only** because pytest dragged it into the production resolution — the comment added in #289 said exactly that, and this is the case it anticipated.

## Verification

`pip install --dry-run --ignore-installed` on both files:

| | Before | After |
|---|---|---|
| Production packages | 62 | **51** |
| Dev packages | 135 | 135 |

- Eleven leave production: `pytest`, `pytest-asyncio`, `pytest-cov`, `pytest-xdist`, `pytest-json-report`, `pytest-metadata`, `pluggy`, `iniconfig`, `execnet`, `coverage`, `pygments`.
- **No new package**, and dev loses nothing and changes no version.
- None of the eleven is imported under `src/`, `main.py`, `alembic/` or `scripts/`.
- The pins stay audited: `pip-audit` and `safety` scan the dev-installed environment, so `pytest==9.0.3` (CVE-2025-71176) and `pygments==2.20.0` (CVE-2026-4539) are still covered.

Taken with #289, the production image goes from **100 packages to 51**.

## Two things the move broke, fixed here

Both came out of the local review, and both are regressions this change would have introduced:

- **`README.md:576`** installed `requirements.txt` in the development setup and then, a few lines below, told the reader to run `pytest tests/unit/ -v`. A fresh clone following it verbatim would have hit `pytest: command not found`. It now installs `requirements-dev.txt`, which begins with `-r requirements.txt` anyway.
- **`ADR-003:125`** still recorded `requirements.txt` as the home of testing dependencies — the kind of stale line that invites someone to put pytest back where it was.

## Also

`pytest.ini` and `dev_tests.py` were still reaching the image through `COPY . .`, which left the stated goal half done. Added to `.dockerignore`.

`pytest-cov` moves **unpinned**, exactly as it was in `requirements.txt` — the only unversioned line in the file. Pinning it is worth doing, but not mixed into a move: it would change a resolution while claiming not to.

## Summary by CodeRabbit

- **Documentation**
  - Updated development setup instructions to install the complete development dependency set.
  - Updated testing strategy documentation to reference the dedicated development requirements file.

- **Chores**
  - Separated testing tools from production dependencies, keeping test tooling out of production installations.
  - Added pinned testing dependencies, including coverage, parallel execution, and reporting tools.
  - Excluded selected test configuration files from Docker build contexts.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -30,6 +30,8 @@ RENDER_DEPLOYMENT.md
 
 # Tests
 tests/
+pytest.ini
+dev_tests.py
 .pytest_cache
 htmlcov/
 .coverage
--- a/README.md
+++ b/README.md
@@ -572,8 +572,8 @@ cd RyderCupAM
 python3.12 -m venv .venv
 source .venv/bin/activate
 
-# Install dependencies
-pip install -r requirements.txt
+# Install dependencies (dev includes production: it starts with -r requirements.txt)
+pip install -r requirements-dev.txt
 
 # Copy environment template
 cp .env.example .env
--- a/docs/architecture/decisions/ADR-003-testing-strategy.md
+++ b/docs/architecture/decisions/ADR-003-testing-strategy.md
@@ -122,4 +122,4 @@ tests/
 - [pytest-xdist Plugin](https://pytest-xdist.readthedocs.io/)
 - [Testing Best Practices](https://testing.googleblog.com/)
 - [Clean Architecture Testing](https://blog.cleancoder.com/uncle-bob/2014/05/14/TheLittleMocker.html)  
-- `requirements.txt`: Documented testing dependencies
+- `requirements-dev.txt`: Documented testing dependencies (moved out of `requirements.txt` in #291, so the production image does not ship the test runner)
--- a/docs/architecture/decisions/ADR-003-testing-strategy.md
+++ b/docs/architecture/decisions/ADR-003-testing-strategy.md
@@ -30,7 +30,7 @@ We need to establish a testing strategy that is:
 **We implement Optimized Testing** with:
 
 ### 1. Framework and Tools:
-- **pytest 8.3.0**: Main framework
+- **pytest 9.0.3**: Main framework
 - **pytest-xdist 3.8.0**: Automatic parallelization
 - **httpx 0.27.0**: HTTP client for API tests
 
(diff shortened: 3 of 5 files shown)


agustinEDev:
@coderabbitai review

